PR #169, opened 2021-09-28
Docker Release

milt:
[SQL-121] Prepare for public Docker Release

Please review the docs page here: https://github.com/yetanalytics/lrsql/blob/docker_release/doc/docker.md

Note that the current image is pushed to `yetanalytics/lrsql:latest` so you can follow along with the doc instructions.

[SQL-121]: https://yet.atlassian.net/browse/SQL-121

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ For releases and release notes, see the [Releases](https://github.com/yetanalyti
 - [Getting Started](doc/startup.md)
 - [Setting up TLS/HTTPS](doc/https.md)
 - [Authority Configuration](doc/authority.md)
+- [Docker Image](docker.md)
 
 ### DBMS-specific Sections
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ For releases and release notes, see the [Releases](https://github.com/yetanalyti
 - [Getting Started](doc/startup.md)
 - [Setting up TLS/HTTPS](doc/https.md)
 - [Authority Configuration](doc/authority.md)
-- [Docker Image](docker.md)
+- [Docker Image](doc/docker.md)
 
 ### DBMS-specific Sections
 
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -2,7 +2,7 @@
 
 # Docker Image
 
-Yet Analytics publishes Docker container images of SQL LRS on [DockerHub](https://hub.docker.com/repository/docker/yetanalytics/lrsql).
+Yet Analytics publishes Docker container images of SQL LRS on [DockerHub](https://hub.docker.com/repository/docker/yetanalytics/lrsql) in the format `yetanalytics/lrsql:<release version | latest>`.
 
 ## Usage
 
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -47,6 +47,8 @@ docker run \
     /lrsql/bin/run_postgres.sh
 ```
 
+Note that you will also need to ensure that a PostgreSQL instance is accessible to the container for this command to work.
+
 ### Customization
 
 The SQL LRS image can be used as a base image for a customized docker image.
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -9,7 +9,6 @@ Yet Analytics publishes Docker container images of SQL LRS on [DockerHub](https:
 You can run SQL LRS directly from the Docker CLI in its default SQLite configuration:
 
 ``` shell
-
 $ docker run \
     -p 8080:8080 \
     -e LRSQL_API_KEY_DEFAULT=my_key \
@@ -17,7 +16,6 @@ $ docker run \
     -e LRSQL_ADMIN_USER_DEFAULT=my_username \
     -e LRSQL_ADMIN_PASS_DEFAULT=my_password \
     yetanalytics/lrsql:latest
-
 ```
 
 ### Other DBMSs
@@ -25,7 +23,6 @@ $ docker run \
 The SQL LRS Dockerfile uses the SQLite run script in `bin` as the default `CMD`. To change to another DBMS use the corresponding script. For instance to use PostgreSQL:
 
 ``` shell
-
 $ docker run \
     -p 8080:8080 \
     -e LRSQL_API_KEY_DEFAULT=my_key \
@@ -40,7 +37,22 @@ $ docker run \
     -e LRSQL_DB_SCHEMA=lrsql \
     yetanalytics/lrsql:latest \
     bin/run_postgres.sh
-
 ```
+
+### Customization
+
+The SQL LRS image can be used as a base image for a customized docker image.
+
+For instance, to make an image with a custom configuration:
+
+``` dockerfile
+FROM yetanalytics/lrsql:latest
+ADD my_lrsql.json /lrsql/config/lrsql.json
+EXPOSE 8080
+EXPOSE 8443
+CMD ["bin/run_postgres.sh"]
+```
+
+The resulting image will use the provided configuration file and run PostgreSQL.
 
 [<- Back to Index](index.md)
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -10,6 +10,7 @@ You can run SQL LRS directly from the Docker CLI in its default SQLite configura
 
 ``` shell
 docker run \
+    -it \
     -p 8080:8080 \
     -e LRSQL_API_KEY_DEFAULT=my_key \
     -e LRSQL_API_SECRET_DEFAULT=my_secret \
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -24,6 +24,7 @@ For releases and release notes, see the [Releases](https://github.com/yetanalyti
 - [Getting Started](doc/startup.md)
 - [Setting up TLS/HTTPS](doc/https.md)
 - [Authority Configuration](doc/authority.md)
+- [Docker Image](docker.md)
```
kelvinqian00, line 27: You forgot the `doc/` prefix here (so when I clicked on it it 404s)

milt: ty

doc/docker.md
```diff
@@ -0,0 +1,64 @@
...
+``` shell
+docker run \
+    -it \
+    -p 8080:8080 \
+    -e LRSQL_API_KEY_DEFAULT=my_key \
+    -e LRSQL_API_SECRET_DEFAULT=my_secret \
+    -e LRSQL_ADMIN_USER_DEFAULT=my_username \
+    -e LRSQL_ADMIN_PASS_DEFAULT=my_password \
+    yetanalytics/lrsql:latest
+```
```
kelvinqian00, line 20: Maybe add a comment saying "You can then communicate with the server, in this example at `0.0.0.0:8080`? I know it's not necessary for someone who knows what they're doing, but I'm speaking as someone who doesn't know what they're doing.

doc/docker.md
```diff
@@ -0,0 +1,64 @@
...
+    -e LRSQL_ADMIN_PASS_DEFAULT=my_password \
+    -e LRSQL_DB_HOST=0.0.0.0 \
+    -e LRSQL_DB_PORT=5432 \
+    -e LRSQL_DB_NAME=lrsql_db \
+    -e LRSQL_DB_USER=lrsql_user \
+    -e LRSQL_DB_PASSWORD=my_password \
+    -e LRSQL_DB_SCHEMA=lrsql \
+    yetanalytics/lrsql:latest \
+    /lrsql/bin/run_postgres.sh
+```
```
kelvinqian00, line 46: Don't you need to setup a running Postgres instance on the image first? I got a `java.net.ConnectException: Connection refused` when I copy-pasted this code in, for obvious reasons. Again, an explanation probably isn't needed for users who know what they're doing, but I definitely think in this case you should show how to prepare the Docker container for PG.

milt: Yep, I feel it might be out of scope to provide the actual step-by-step on that, but a note that you'll need to set up Postgresql first would be good. I leave that up to @cliffcaseyyet actually, if you want a full config then I'll go ahead and do that `compose.yml` file I mentioned.

doc/docker.md
milt: done. I think in general this is intended to be for more advanced users, though it does occur to me that docker is hands-down the _quickest_ way for someone to "try out" lrsql, is it maybe worth linking to this section from getting started?

doc/docker.md
milt: just the note, pending what Cliff wants https://github.com/yetanalytics/lrsql/pull/169/commits/ea1a7c4f5d42cb87b3fc3222383834f0c903abd8

doc/docker.md
```diff
@@ -0,0 +1,64 @@
+[<- Back to Index](index.md)
+
+# Docker Image
+
+Yet Analytics publishes Docker container images of SQL LRS on [DockerHub](https://hub.docker.com/repository/docker/yetanalytics/lrsql) in the format `yetanalytics/lrsql:<release version | latest>`.
+
+## Usage
```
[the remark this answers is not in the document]
milt: done, with current nesting preserved

doc/docker.md
milt: @kelvinqian00 see https://github.com/yetanalytics/lrsql/pull/169/commits/31519d88b723401886c0e4593e63dfa6e8d991d6

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+      - db
+    environment:
+      LRSQL_API_KEY_DEFAULT: my_key
+      LRSQL_API_SECRET_DEFAULT: my_secret
+      LRSQL_ADMIN_USER_DEFAULT: my_username
+      LRSQL_ADMIN_PASS_DEFAULT: my_password
+      LRSQL_DB_HOST: db
+      LRSQL_DB_NAME: lrsql_db
+      LRSQL_DB_USER: my_db_user
+      LRSQL_DB_PASSWORD: my_db_password
```
kelvinqian00, line 31: The env vars here (except for `LRSQL_DB_NAME`) differ from the ones in the doc. Also, the env vars here do not set a schema, whereas the ones in the doc do.

milt: The docker example is not the compose file. They are different ways to launch.

As for the env vars I can make them match except for LRSQL_DB_HOST - different because that's how hostnames work. In Compose `db` is the hostname of the database service (look above in the file)

For the schema, we cannot set our own without additional scripting to create the schema. If we omit it everything works fine, so I do!

kelvinqian00: Yeah, I want the doc example and the compose file to line up as much as possible. Obviously I don't want them to line up to the point it breaks everything.

In the schema case, it's probably better to omit the schema from the doc example than to keep it - less complexity overall in addition to lining up with the compose file.

milt: See https://github.com/yetanalytics/lrsql/pull/169/commits/c97bcd9f6d85cfee15d13bf03303013c34331302

That's as lined up as they get without breaking. I removed schema.

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    # build: . # switch to this for active dev
+    image: yetanalytics/lrsql:latest
+    command:
+      - /lrsql/bin/run_postgres.sh
+    ports:
+      - "8080:8080"
+    depends_on:
+      - db
+    environment:
+      LRSQL_API_KEY_DEFAULT: my_key
```
kelvinqian00, line 24: Based on talks between Cliff and I, adding `LRSQL_POOL_INITIALIZATION_FAIL_TIMEOUT: 10000` here should suffice for avoiding connection failures.

doc/docker.md
```diff
@@ -0,0 +1,73 @@
...
+    -e LRSQL_DB_HOST=0.0.0.0 \
+    -e LRSQL_DB_PORT=5432 \
+    -e LRSQL_DB_NAME=lrsql_db \
+    -e LRSQL_DB_USER=lrsql_user \
+    -e LRSQL_DB_PASSWORD=lrsql_password \
+    yetanalytics/lrsql:latest \
+    /lrsql/bin/run_postgres.sh
+```
+
+Note that you will also need to ensure that a PostgreSQL instance is accessible to the container for this command to work. For a demonstration of containerized SQL LRS with PostgreSQL you can use the `docker-compose.yml` file in the project root:
```
kelvinqian00, line 49: I noticed that you use "PostgreSQL" as opposed to "Postgres" here. I actually made all the documentation say "Postgres" for consistency reasons, so may want to continue that.

kelvinqian00 (approved):
image

doc/docker.md
```diff
@@ -0,0 +1,73 @@
...
+    /lrsql/bin/run_postgres.sh
+```
+
+Note that you will also need to ensure that a Postgres instance is accessible to the container for this command to work. For a demonstration of containerized SQL LRS with Postgres you can use the `docker-compose.yml` file in the project root:
+
+``` shell
+docker compose up
+```
+
+Docker will start Postgres and then SQL LRS. Note that Postgres can sometimes take a while to start causing SQL LRS initialization to fail. If this happens, stop the system with ctrl-C and run the command again.
```
kelvinqian00, line 55: Oops spoke a bit too soon - we could remove or revise this.

cliffcaseyyet (commented):
dig the pass-through env approach. just sidesteps everything. everything looks good but one question - is there a feasible path to PEM or keystore injection, or to authority config?

milt:
> dig the pass-through env approach. just sidesteps everything. everything looks good but one question - is there a feasible path to PEM or keystore injection, or to authority config?

See https://github.com/yetanalytics/lrsql/blob/docker_release/doc/docker.md#customization

kelvinqian00 (approved):
image

cliffcaseyyet (approved):

doc/docker.md
```diff
@@ -0,0 +1,84 @@
...
+# custom configuration
+ADD my_lrsql.json              /lrsql/config/lrsql.json
+
+# custom certs
+ADD my_server.key.pem          /lrsql/config/server.key.pem
+ADD my_server.crt.pem          /lrsql/config/server.crt.pem
+ADD my_cacert.pem              /lrsql/config/cacert.pem
+
+# custom authority
+ADD my_authority.json.template /lrsql/config/authority.json.template
```
kelvinqian00, line 75: I noticed that these are all prefixed with `my_`, which might require additional env var overrides? Probably simpler to go with the defaults.

milt: They won't, because they are going to the default file locations.

milt: like, look at the right side of the `ADD` directives

kelvinqian00: My bad, I was reading them from left-to-right.

PR #174, opened 2021-09-29
Change docker badge color from orange to blue

kelvinqian00:
Make the Docker badge more consistent with the Docker(hub) color scheme, and also because blue is a nicer color than orange in my opinion.


milt (approved):
_In my dreams... So beautiful..._

milt:
Fremen Button

PR #175, opened 2021-09-29
contributor covenant badge

cliffcaseyyet:
since we're doing badges...


kelvinqian00:
Is it possible to change the badge color to purple? Simply because the contributor covenant webpage uses purple in its header.

cliffcaseyyet:
> Is it possible to change the badge color to purple? Simply because the contributor covenant webpage uses purple in its header.

surprisingly....yes

kelvinqian00 (approved):

PR #176, opened 2021-09-30
Add persistent Sqlite volume to Docker example.

milt:
It occurred to me that the simple Docker example we give is effectively ephemeral, since the data is only written within the container. This PR modifies the example run command to include a bind mount so data will persist. A named volume is used so we don't have to worry about absolute paths, and the example "just works".

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -16,10 +16,14 @@ docker run \
     -e LRSQL_API_SECRET_DEFAULT=my_secret \
     -e LRSQL_ADMIN_USER_DEFAULT=my_username \
     -e LRSQL_ADMIN_PASS_DEFAULT=my_password \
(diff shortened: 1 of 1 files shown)


kelvinqian00:
Do the Postgres instructions have to be updated as well or do they already make a persistent DB?

milt:
> Do the Postgres instructions have to be updated as well or do they already make a persistent DB?

They persist with the Compose instructions, but the raw postgres invocation, no, it would be outside its purview (connecting to a pg instance somewhere)

kelvinqian00 (approved):

cliffcaseyyet (approved):
I dig it. And I prefer volumes being the canonical way to inject config. I think this method allows them to update new versions we publish in-place without creating a new custom image. I still like the other approach being documented.

PR #186, opened 2021-11-22
[SQL-126] Change `url-prefix` to `path-prefix`

kelvinqian00:
`add-xapi-interceptors` accepts `path-prefix`, not `url-prefix` as the arg to change the URL path prefix (e.g. `host:port/xapi`).


milt (approved):
👍

milt (approved):
That is a lot clearer, looks good.

PR #189, opened 2021-12-10
[SQL-125] Redact sensitive information from traces

kelvinqian00:
- Redact sensitive info before logging config var maps
- ~Turn off Pedestal logging since we can't redact info from that~ (Made redundant by https://github.com/yetanalytics/lrs/pull/71)
- Return 400 instead of 500 error (which spews out exn info) upon JSONParseException
- ~TODO: Write tests for the above~ Done!

### DIFF
--- a/src/main/lrsql/system/database.clj
+++ b/src/main/lrsql/system/database.clj
@@ -3,7 +3,7 @@
             [com.stuartsierra.component :as component]
             [lrsql.backend.protocol :as bp]
             [lrsql.spec.config :as cs]
-            [lrsql.system.util :refer [assert-config make-jdbc-url]])
+            [lrsql.system.util :as cu :refer [assert-config make-jdbc-url]])
   (:import [com.zaxxer.hikari HikariConfig HikariDataSource]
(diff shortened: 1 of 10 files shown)


src/main/lrsql/system/database.clj
```diff
@@ -97,7 +97,7 @@
       (if-not ?conn-pool
         (let [conn-pool (make-conn-pool backend config)]
           (log/infof "Starting new connection for %s database..." db-type)
-          (log/debugf "Config: %s" config)
+          (log/debugf "Config: %s" (cu/redact-config-vars config))
```
milt, line 100: After reading the `c.t.l` source, the walk is fine. If this only happens in debug, the expanded macro contains a conditional on level.

milt (approved):
Looks sensible enough to me

cliffcaseyyet (approved):
nice

milt (approved):
Yep, looks good. As expected the json parse catcher after the general error handler will get called before it on the exit interceptor pass.

cliffcaseyyet (approved):

PR #190, opened 2021-12-10
sql-128 - limiting log archives to 60d

cliffcaseyyet:


kelvinqian00 (approved):
Looks good to me (and I trust you that it works).

PR #191, opened 2021-12-16
Add nvd-clojure support and update deps

kelvinqian00:
- [nvd-clojure](https://github.com/rm-hull/nvd-clojure) support:
  - Add `:nvd` alias in `deps.edn`
  - Add `check-vuln` make target
  - NOTE: Have not added CI integration since that is much more involved
- Update deps to address CVEs
  - logback to 1.2.8
  - cheshire to 5.10.1
  - ring-core to 1.9.4
  - jetty deps to 9.4.44
  - NOTE: It is cleaner to update Pedestal instead of the ring and jetty deps individually
  - NOTE: Have not updated H2 to v2.0.x since that is a major version change


Makefile
```diff
@@ -73,6 +73,13 @@ bench-async:
 		-a true \
 		-u username -p password
 
+# Vulnerability check
+
+target/nvd:
+	clojure -Xnvd check :classpath '"'"$$(clojure -Spath -A:db-h2:db-sqlite:db-postgres)"'"'
+
+check-vuln: target/nvd
```
milt, line 81: By default `nvd-clojure` just outputs to `target/nvd` so you don't need a phony here really

milt: oh nm I see what you did here

milt (approved):
This looks good. I've yet to get a satisfying solution for CI, currently investigating how to specifically cache + hash it

kelvinqian00:
After thinking about it, I decide it's best to update these relatively modest changes now rather than wait for everything:
- Pedestal does not seem like it will get updated soon, so the temp Ring/Jetty solution is the best we have at the moment.
- If we add `check-vuln` to CI, it will fail until H2 is updated to v2.x.x, which is would be a lot.

cliffcaseyyet (approved):
looks good.

PR #192, opened 2021-12-20
SQL-130 - Markdoc conversion

cliffcaseyyet:
converted to markdoc dep, 
removed doc ns. 
Still need final sha before approval (pending markdoc PR)


Makefile
```diff
@@ -107,8 +107,7 @@ target/bundle/bin:
 # Create HTML docs
 
 target/bundle/doc:
-	clojure -M:doc -m lrsql.render-doc doc target/bundle/doc
-	cp -r doc/images/ target/bundle/doc/images/
```
kelvinqian00, line 111: Don't we still need to copy the images over? Or does the new lib do that for us?

kelvinqian00: Ok downloaded the artifact they do get copied over.

kelvinqian00 (approved):

PR #193, opened 2021-12-27
It'll be 2022 soon

kelvinqian00:


cliffcaseyyet (approved):
Obviously this is fine for now but i do wish there was a way in MD to handle "current year". The solutions i saw on SO were pretty sketchy, like a dynamic SVG or something. Also you may want to bundle this with a UI update. The year is in one spot in footer, and in that case actually _can_ be dynamic. (https://github.com/yetanalytics/lrs-admin-ui/blob/main/src/com/yetanalytics/lrs_admin_ui/views/footer.cljs)

milt:
> The solutions i saw on SO were pretty sketchy, like a dynamic SVG or something.

I wonder if shields.io has something for this, we've already accepted it into the fold..

PR #195, opened 2021-12-29
updated module list. upgraded runtimer usage to reusable workflow

cliffcaseyyet:
- Added crypto module to java modules for custom JREs because of SSL issue
- Added `.java_modules` list
- Updated runtimer usage to use re-usable workflow
- Switched to locally generated runtime for dev
- Updated Dockerfile to use modules list


milt (approved):
This looks good to me, incl docker

PR #197, opened 2021-12-29
Add NVD scanning to SQL LRS workflow

kelvinqian00:


milt:
Don't forget a scheduled one too when you come back to this.

kelvinqian00:
> Don't forget a scheduled one too when you come back to this.

That was what I did in https://github.com/yetanalytics/lrsql/pull/197/commits/2a86d797a94c724eb9b7a8f78f40da11c0e85786.

kelvinqian00:
^ Going to keep the local alias + Make target around in order to make local dev easy.

.github/workflows/nvd.yml
```diff
@@ -0,0 +1,12 @@
+name: NVD Scan
```
milt, line 1: You should specify here (if not in the filename) that this is a schedule scan

deps.edn
```diff
@@ -133,7 +133,7 @@
   {:extra-deps  {com.github.seancorfield/depstar {:mvn/version "2.1.267"}}
    :extra-paths ["src/build"]} 
   :nvd
-  {:replace-deps {nvd-clojure/nvd-clojure {:mvn/version "1.9.0"}}
+  {:replace-deps {nvd-clojure/nvd-clojure {:mvn/version "2.0.0"}}
```
milt, line 136: I believe the scanner provides nvd itself, so no need to have this here (unless there is some other call in the makefile that needs it)

milt (approved):
Looks good, see minor notes

deps.edn
kelvinqian00: It's for the `check-vuln` Makefile target. Neither this nor `check-vuln` are strictly necessary anymore, but as I said above I want to keep it for easy manual dev. Another way would be to get rid of this alias but keep `check-vuln`, having the latter use clojure.tools, but that would require the _user_ to ensure that they have nvm-clojure installed.

.github/workflows/nvd.yml
kelvinqian00: https://github.com/yetanalytics/lrsql/pull/197/commits/fea48c7d6e6af3eacdee27cad067f6ca83fc8d90

PR #200, opened 2022-01-27
SQL-132 remove clojure.main from all invocations

milt:
[SQL-132] For some reason (maybe the old depstar docs?) we always invoke lrsql through clojure.main, but all of our entrypoints are AOT'd so this is superfluous. Removing them simplifies things and likely gives a little speed-up to startup.

This should also apply to the exe's, but someone with windows will have to tell me if that works ;)

### DIFF
--- a/bin/run_h2.sh
+++ b/bin/run_h2.sh
@@ -2,4 +2,4 @@
 
 MACHINE=`bin/machine.sh`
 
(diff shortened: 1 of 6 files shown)


milt:
NB: Milt had an off-by-one error on the branch name, be aware if of this if you want to run the branch docker, for instance

kelvinqian00 (approved):
Looks good, but my approval is conditional on testing on Windows.

milt:
Wandoz confirmed 
image

PR #201, opened 2022-02-03
Fix SQL syntax

rajkowski:


kelvinqian00 (approved):
Thank you for catching the missing semicolons! We really appreciate this sort of feedback from the community.

PR #202, opened 2022-02-08
Allow configuration of tc container version

milt:
This shows the jdbc uri method for specifying container version. I don't feel strongly that it should be implemented the way it is here, could be passed in the config map itself, just a demonstration.

### DIFF
--- a/src/test/lrsql/test_support.clj
+++ b/src/test/lrsql/test_support.clj
@@ -89,7 +89,8 @@
 ;; `tc`-prefixed DB types.
 
 (defn fresh-postgres-fixture
-  [f]
+  [f & {:keys [version]
+        :or {version "latest"}}]
   (let [id-str (str (UUID/randomUUID))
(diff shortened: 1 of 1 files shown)


kelvinqian00:
Hmm, the main difference between `version` and any other configurable value is that it's test-only, so it actually makes some sort of sense to inject that value this way. Though with Aero you can easily have config values that only exist in particular profiles; we'd need a way to discourage users from setting versions in production though (unless there's some way to make that useful in prod?).

milt:
> we'd need a way to discourage users from setting versions in production

I think this is a case where the "open" nature of map keys make it fine to have a test-only value in the config, assuming nothing causes differences in prod (the fixture is only run in test).

But we could also just consider it part of the `:db-name` and just directly put in `tc:postgresql:<version>` from the config, right?

milt:
Oh I see, `next.jdbc` doesn't like it

PR #203, opened 2022-02-08
add tc version via config

milt:
This PR adds a key to the `:database` config called `:test-db-version` which will be used (in tests ONLY) to set the version of PG testcontainers provides for fixtures.

### DIFF
--- a/resources/lrsql/config/config.edn
+++ b/resources/lrsql/config/config.edn
@@ -22,7 +22,7 @@
                    ;; :db-properties "currentSchema=lrsql"
 
                    ;; Testing Only! Specify the version used with testcontainers
-                   :test-db-version "latest"}
+                   :test-db-version "11.9"}
    ;; Production
    :prod-h2-mem   #include "prod/h2_mem/database.edn"
    :prod-h2       #include "prod/h2/database.edn"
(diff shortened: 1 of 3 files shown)


kelvinqian00:
This definitely looks cleaner than the other way. Now I wonder if it should be added to the config spec (which I'm leaning towards no since the config spec is more for validation during production).

milt:
> if it should be added to the config spec

milt:
Actually, we could put test stuff like this off in its own top-level key, might be even cleaner?

kelvinqian00:
The issue with that is that this config value isn't just test-specific, it's also _Postgres_-specific. So I actually don't think it would be cleaner, since now you have PG-specific config leaking up to the top level.

milt:
Oh, I meant more like a `:test` key in the config at the top, with stuff like db versions underneath, but in the end I don't really care too much one way or the other

milt:
I think the prefix of `:test-` and the comments are sufficient to let folks know it has no prod effect

kelvinqian00 (approved):
I suggest adding a comment in the config schema but otherwise looks good.

milt:
For instance, if we wired the config to a testing-only env var we could make a gh actions matrix and explicitly test certain versions, etc

PR #204, opened 2022-02-16
SQL-133 OIDC/JWKS Config

milt:
[SQL-133] Add config for unsigning foreign tokens

### DIFF
--- a/resources/lrsql/config/prod/default/webserver.edn
+++ b/resources/lrsql/config/prod/default/webserver.edn
@@ -13,4 +13,7 @@
  :ssl-port          #long #or [#env LRSQL_SSL_PORT 8443]
  :url-prefix        #or [#env LRSQL_URL_PREFIX "/xapi"]
  :enable-admin-ui   #boolean #or [#env LRSQL_ENABLE_ADMIN_UI true]
- :enable-stmt-html  #boolean #or [#env LRSQL_ENABLE_STMT_HTML true]}
+ :enable-stmt-html  #boolean #or [#env LRSQL_ENABLE_STMT_HTML true]
+ :openid-issuer     #env LRSQL_OPENID_ISSUER
+ :openid-config     #env LRSQL_OPENID_CONFIG
+ :jwks-uri          #env LRSQL_JWKS_URI}
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -110,5 +110,8 @@ You may have noted that some options are not available:
 | `LRSQL_URL_PREFIX` | `urlPrefix` | The prefix of the webserver URL path, e.g. the prefix in `http://0.0.0.0:8080/xapi` is `/xapi`. Used when constructing the `more` value for multi-statement queries. | `/xapi` |
 | `LRSQL_ENABLE_ADMIN_UI` | `enableAdminUi` | Whether or not to serve the administrative UI at `/admin` | `true` |
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
+| `LRSQL_OPENID_ISSUER` | `openidIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
+| `LRSQL_OPENID_CONFIG` | `openidConfig` | Custom OIDC configuration location | Not set |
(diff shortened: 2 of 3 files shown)


resources/lrsql/config/prod/default/webserver.edn
```diff
@@ -13,4 +13,7 @@
  :ssl-port          #long #or [#env LRSQL_SSL_PORT 8443]
  :url-prefix        #or [#env LRSQL_URL_PREFIX "/xapi"]
  :enable-admin-ui   #boolean #or [#env LRSQL_ENABLE_ADMIN_UI true]
- :enable-stmt-html  #boolean #or [#env LRSQL_ENABLE_STMT_HTML true]}
+ :enable-stmt-html  #boolean #or [#env LRSQL_ENABLE_STMT_HTML true]
+ :openid-issuer     #env LRSQL_OPENID_ISSUER
+ :openid-config     #env LRSQL_OPENID_CONFIG
+ :jwks-uri          #env LRSQL_JWKS_URI}
```
kelvinqian00, line 19: Hmm I guess I never realized you could use `#env` without `#or`. Do these values get assigned `nil` or do they just not appear in the config map (which would have implications for the config spec)?

milt: It was my impression that they come through as nil, but then a function cleans out nils. I'll double check that now

kelvinqian00: Oh yeah I forgot I added that function which means the spec should work. But double checking is always sensible.

milt: Ah, though there are spec conformers, they don't necessarily get run that I can see. BUT it looks like `aero` actually omits the key entirely if it is just env and isn't set.

kelvinqian00: Dang I wish I knew that earlier.

milt: Well I don't see that it's documented, I'd assume the same you did. But the repl does indicate this is the case

src/main/lrsql/util/oidc.clj
```diff
@@ -0,0 +1,20 @@
+(ns lrsql.util.oidc
+  "OpenID Connect Utilities"
+  (:require [clojure.spec.alpha :as s]
+            [clojure.string :as cs]
+            [com.yetanalytics.pedestal-oidc.discovery :as disco]
+            [lrsql.spec.config :as config]))
```
kelvinqian00, line 6: Looks like `clojure.string` and `lrsql.spec.config` are unused in this namespace.

src/test/lrsql/util/oidc_test.clj
```diff
@@ -0,0 +1,16 @@
+(ns lrsql.util.oidc-test
+  (:require [clojure.test :refer :all]
+            [lrsql.util.oidc :refer :all]))
```
kelvinqian00, line 3: Oh dear VSCode is going to scream at these `:all`s.

milt: Why? Is this a kondo thing? That is practically idiomatic in clojure tests

src/main/lrsql/util/oidc.clj
```diff
@@ -0,0 +1,20 @@
...
+
+(s/fdef get-configuration
+  :args (s/cat :webserver ::config/webserver)
+  :ret (s/nilable map?))
+
+(defn get-configuration
```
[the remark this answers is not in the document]
milt: In the next bit of this I write (note that this is not a PR to master, but to a feature branch). Basically you'd use that for general OIDC discovery, I'm just doing a specific slice of that next, the keyset.

src/test/lrsql/util/oidc_test.clj
kelvinqian00: It is indeed a kondo thing. Granted you can suppress it using `#_{:clj-kondo/ignore [:refer-all]}` but otherwise it'll be annoying.

kelvinqian00 (approved):

dev-resources/oidc/.well-known/openid-configuration
```diff
@@ -0,0 +1,106 @@
+{
```
cliffcaseyyet, line 1: wait why does it need a discovery address resource? Is this like application-as-resource type thing?

milt: Nope, this is just for tests. I'm punning on the fact that it treats file locs + uris interchangeably

milt: see https://github.com/yetanalytics/lrsql/blob/e819ec40dfb68a054351d11bd5bad2d83511b8aa/src/test/lrsql/util/oidc_test.clj

cliffcaseyyet: gotcha, ok cool.

PR #205, opened 2022-02-16
SQL-135 SQL-134 OIDC token auth 

milt:
[SQL-135][SQL-134] Enable token authn + authz with OIDC OAuth 2.0.

This PR enables OIDC for xapi authn/authz ONLY, and does not address the Admin API/UI.

To try it out with keycloak, use this branch of `re-oidc` dev env: https://github.com/yetanalytics/re-oidc/tree/lrsql_testing and run:

    LRSQL_HTTP_PORT=8081 \
    LRSQL_OIDC_ISSUER=http://0.0.0.0:8080/auth/realms/test \
    make ephemeral-prod

To try it out with Auth0, use https://github.com/yetanalytics/re-oidc/tree/lrsql_testing_auth0 and run:

    LRSQL_HTTP_PORT=8081 \
    LRSQL_OIDC_ISSUER=https://milt.auth0.com \
    make ephemeral-prod

[SQL-135]: https://yet.atlassian.net/browse/SQL-135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SQL-134]: https://yet.atlassian.net/browse/SQL-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/lrsql/util/oidc.clj
```diff
@@ -2,18 +2,71 @@
...
+          (fn [_]
+            (fn [kid]
+              (get @keyset-cache kid
+                   ;; If kid is not found in the keyset, attempt refresh and try
+                   ;; again
+                   (get (reset! keyset-cache (jwt/get-keyset jwks-uri))
+                        kid))))
+          :required? false)
+         ;; TODO: parse out scopes + authorize
+         ])
```
milt, line 63: @cliffcaseyyet kinda split the baby here with the keyset cache thingy, keeping it scoped but not introducing a lot of code to make it a first-class component.

src/main/lrsql/util/oidc.clj
```diff
@@ -1,19 +1,120 @@
...
+                               not-empty
+                               (->> (into #{})))]
+         (let [;; Roughly following https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#oauth-credentials-as-authority
+               authority {"objectType" "Group"
+                          "member"     [{"account"
+                                         {"homePage" iss
+                                          "name"     azp}}
+                                        {"account"
+                                         {"homePage" iss
+                                          "name"     sub}}]}]
```
milt, line 114: There isn't really a well defined way to do this in the spec for pure token claims, so I've done an approximation

src/main/lrsql/system/lrs.clj
```diff
@@ -28,7 +28,8 @@
             [lrsql.util.auth      :as auth-util]
             [lrsql.util.statement :as stmt-util]
             [lrsql.init.authority   :refer [make-authority-fn]]
-            [lrsql.util.concurrency :refer [with-rerunable-txn]])
+            [lrsql.util.concurrency :refer [with-rerunable-txn]]
+            [lrsql.util.oidc :as oidc])
```
kelvinqian00, line 32: I wonder how much of the stuff in `lrsql.util.oidc` can be split off into a `lrsql.init.oidc` namespace.

src/main/lrsql/util/oidc.clj
```diff
@@ -1,19 +1,187 @@
 (ns lrsql.util.oidc
   "OpenID Connect Utilities"
```
kelvinqian00, line 2: Going off of the above and looking more closely at the functions in this namespace, I think `get-configuration`, `resource-interceptors`, `resolve-authority-claims`, and `make-authority-fn` could be moved to a `lrsql.init.oidc` namespace, leaving `parse-scope-claim` and `token-auth-identity` in this namespace.

src/main/lrsql/system/lrs.clj
milt: I tried to keep the footprint as small as possible. What would go there, the authority function stuff?

src/main/lrsql/util/oidc.clj
```diff
@@ -1,19 +1,187 @@
...
+         ;; no valid scopes, can't do anything
+         :com.yetanalytics.lrs.auth/unauthorized)})))
+
+;; Authority
+
+(s/fdef resolve-authority-claims
+  :args (s/cat :claims ::oidc/claims)
+  :ret ::oidc/authority-claims)
+
+(defn resolve-authority-claims
```
kelvinqian00, line 142: I'm on the fence on whether `get-configuration` or `resolve-authority-claims` should be private functions or not. On one hand they're only used by other functions in this namespace, but I _could_ see them called from other namespaces in the future. (And they certainly need to be public for tests!)

src/main/lrsql/util/oidc.clj
```diff
@@ -1,19 +1,187 @@
...
-  via :openid-issuer or :openid-config."
-  [{:keys [openid-issuer
-           openid-config]}]
-  (when-let [config-uri (or openid-config
-                            (and openid-issuer
-                                 (disco/issuer->config-uri openid-issuer)))]
-    (disco/get-openid-config config-uri)))
+  via :oidc-issuer or :oidc-config."
+  [{:keys               [oidc-issuer
+           oidc-config] :as config}]
```
kelvinqian00, line 37: Weird formatting?

src/main/lrsql/util/oidc.clj
milt: Sure, that sounds fine

src/main/lrsql/util/oidc.clj
```diff
@@ -1,19 +1,187 @@
...
-            [lrsql.spec.config :as config]))
+            [com.yetanalytics.pedestal-oidc.interceptor :as oidc-i]
+            [com.yetanalytics.pedestal-oidc.jwt :as jwt]
+            [io.pedestal.interceptor :as i]
+            [lrsql.init.authority :as authority]
+            [lrsql.spec.config :as config]
+            [lrsql.spec.oidc :as oidc]
+            [lrsql.util.auth :as auth]
+            [selmer.parser :as selm-parser]
+            [xapi-schema.spec :as xs])
```
kelvinqian00, line 18: `xapi-schema.spec` is an unused require.

src/main/lrsql/util/oidc.clj
```diff
@@ -1,19 +1,187 @@
...
+        ex)))))
+
+(s/fdef parse-scope-claim
+  :args (s/cat :scope-str string?)
+  :ret (s/every ::auth/scope))
+
+(defn parse-scope-claim
+  [scope-str]
+  (keep auth/scope-str->kw
+        (cs/split scope-str #"\s")))
```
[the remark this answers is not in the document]
milt: This might actually get more complex: Currently the scope names are very general (ex. `all`) and we're probably going to want to allow a configurable prefix as many IDP scenarios consider scopes global, probably leading to a need for tests, in which case it could not be private.

src/main/lrsql/util/oidc.clj
milt: haha clj-fmt edge case

src/main/lrsql/util/oidc.clj
milt: Nope, the `:statement/authority` key is defined there

src/main/lrsql/system/lrs.clj
kelvinqian00: Yes, as per my other comment. And I don't think two namespaces is much more of a concern than one, especially when you have this clear divide between "init" and "runtime".

src/main/lrsql/util/oidc.clj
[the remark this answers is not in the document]
kelvinqian00: In this case there should be a comment explaining why this helper function should be public.

Makefile
```diff
@@ -47,6 +47,10 @@ ci: test-h2 test-sqlite test-postgres
 ephemeral: resources/public/admin
 	clojure -X:db-h2 lrsql.h2.main/run-test-h2 :persistent? false
 
+# like ephemeral, but takes env vars
+ephemeral-prod: resources/public/admin
+	clojure -M:db-h2 -m lrsql.h2.main --persistent false
```
kelvinqian00, line 52: Have you tried using the regular `ephemeral` version and just add env-var-supporting vars to the `:test` aero profile?

milt: The idea here is that it is a quick way to launch the application _as it will be offered in the distribution_, so I don't see the utility of adding them statically like that. This lets me pass env vars from the cli without editing source.

src/main/lrsql/util/oidc.clj
milt: see https://github.com/yetanalytics/lrsql/pull/205/commits/50247e22222e58a8d11cbb94e804a599dbc17a0a

src/main/lrsql/util/oidc.clj
milt: fixed in https://github.com/yetanalytics/lrsql/pull/205/commits/50247e22222e58a8d11cbb94e804a599dbc17a0a

src/main/lrsql/util/oidc.clj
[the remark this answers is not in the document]
milt: I think I'll just implement the rest of the scope prefix feature, we're almost certainly gonna need it anyhow, and it makes this function less trivial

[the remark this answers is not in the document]
milt: Implemented the scope prefix here: https://github.com/yetanalytics/lrsql/pull/205/commits/0002110e9983fcad6d220ae001185494b8ccd817

kelvinqian00:
Also don't forget to add the dev instructions to the documentation.

milt:
> Also don't forget to add the dev instructions to the documentation.

My plan was to do these comprehensively as another PR to the [feature branch](https://github.com/yetanalytics/lrsql/tree/SQL-132). That includes the dev info on how to try/use an IDP from the CLI, but also the full customer-facing instructions. I'm quite sure both of those will have revisions as I refine this more and hook it up to admin stuff, hence the separation.

resources/lrsql/config/oidc_authority.json.template
```diff
@@ -0,0 +1,27 @@
+{# for template syntax see: https://github.com/yogthos/Selmer#templates
+
+   Use any variable present in token claims
+   https://datatracker.ietf.org/doc/html/rfc9068
+
+   Special Variables:
+
+   * lrsql/resolved-client-id - Replacement for aud that is always a string.
```
kelvinqian00, line 8: Maybe document here that `iss` and `sub` are fields that will come from the JWT.

milt: That link leads to a master list of everything that could (possibly) be in the claims with definitions, isn't that enough?

milt: oh, that's not actually the doc I thought I linked there

milt: see https://github.com/yetanalytics/lrsql/pull/205/commits/39e22ef25147879a888332da2e6238cdf8c3696a

kelvinqian00 (approved):

PR #206, opened 2022-02-21
[SQL-132] OIDC Support

milt:
[SQL-132] Support OIDC authentication (authn) and authorization(authz) in SQL LRS.

This includes:
- [x] Direct use of the LRS as an OAuth 2.0 resource server, expecting OIDC tokens + claims.
- [x] Authn to the LRS Admin SPA and authn/z to the admin API to manipulate long-lived credentials.
- [x] Implement enable/disable of local account auth when OIDC is used

[SQL-132]: https://yet.atlassian.net/browse/SQL-132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -85,6 +85,8 @@ You may have noted that some options are not available:
 | `LRSQL_STMT_GET_MAX` | `stmtGetMax` | The maximum allowed `limit` value for a statement query. If an explicit `limit` value exceeds this value, it will be overridden. | `50` |
 | `LRSQL_AUTHORITY_TEMPLATE` | `authorityTemplate` | The filepath to the Statement authority template file, which describes how authorities are constructed during statement insertion. If the file is not found, the system defaults to a default authority function. | <details>`config/authority.json.template`<summary>(Filepath)</summary></details> |
 | `LRSQL_AUTHORITY_URL` | `authorityUrl` | The URL that is set as the `authority-url` value when constructing an authority from a template. | `http://example.org` |
+| `LRSQL_OIDC_AUTHORITY_TEMPLATE` | `oidcAuthorityTemplate` | Like `LRSQL_AUTHORITY_TEMPLATE`, but only used when forming an authority from an OIDC access token. | <details>`config/oidc_authority.json.template`<summary>(Filepath)</summary></details> |
+| `LRSQL_OIDC_SCOPE_PREFIX` | `oidcScopePrefix` | An optional prefix prepended to OIDC scope. For example, setting this to `lrs:` would change the expected `all` scope to `lrs:all` | `""` |
 | `LRSQL_STMT_RETRY_LIMIT` | `stmtRetryLimit` | The number of times to retry a statement post transaction before failing. | `10` |
 | `LRSQL_STMT_RETRY_BUDGET` | `stmtRetryBudget` | The max amount of time allowed for statement POST transaction retries before failing (ms). | `1000` |
 
@@ -110,8 +112,8 @@ You may have noted that some options are not available:
 | `LRSQL_URL_PREFIX` | `urlPrefix` | The prefix of the webserver URL path, e.g. the prefix in `http://0.0.0.0:8080/xapi` is `/xapi`. Used when constructing the `more` value for multi-statement queries. | `/xapi` |
 | `LRSQL_ENABLE_ADMIN_UI` | `enableAdminUi` | Whether or not to serve the administrative UI at `/admin` | `true` |
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
-| `LRSQL_OPENID_ISSUER` | `openidIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
-| `LRSQL_OPENID_CONFIG` | `openidConfig` | Custom OIDC configuration location | Not set |
+| `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
+| `LRSQL_OIDC_CONFIG` | `oidcConfig` | Custom OIDC configuration location | Not set |
 | `LRSQL_JWKS_URI` | `jwksUri` | Custom JWKS [keyset](https://datatracker.ietf.org/doc/html/rfc7517#section-5) location | Not set |
 
 [<- Back to Index](index.md)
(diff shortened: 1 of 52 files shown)


doc/env_vars.md
```diff
@@ -110,5 +112,8 @@ You may have noted that some options are not available:
 | `LRSQL_URL_PREFIX` | `urlPrefix` | The prefix of the webserver URL path, e.g. the prefix in `http://0.0.0.0:8080/xapi` is `/xapi`. Used when constructing the `more` value for multi-statement queries. | `/xapi` |
 | `LRSQL_ENABLE_ADMIN_UI` | `enableAdminUi` | Whether or not to serve the administrative UI at `/admin` | `true` |
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
+| `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
+| `LRSQL_OIDC_CONFIG` | `oidcConfig` | Custom OIDC configuration location | Not set |
+| `LRSQL_JWKS_URI` | `jwksUri` | Custom JWKS [keyset](https://datatracker.ietf.org/doc/html/rfc7517#section-5) location | Not set |
```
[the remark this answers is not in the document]
milt: OIDC stuff currently takes effect if any one of the three is present, in the order 1. issuer 2. config 3. jwks_uri. I'm pretty sure I'd like to remove everything but the issuer tho. If they are not provided, no OIDC interceptors are loaded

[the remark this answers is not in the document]
milt: Actually that order is wrong, it is backwards 1. jwks_uri 2. config 3. issuer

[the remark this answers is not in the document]
milt: I'm gonna actually remove the other two (leaving issuer) on https://github.com/yetanalytics/lrsql/pull/207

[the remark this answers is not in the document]
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/5b2bd99061ea81313bb8bf695ec0f0614a6ab8d4

doc/env_vars.md
```diff
@@ -110,5 +112,10 @@ You may have noted that some options are not available:
...
 | `LRSQL_ENABLE_ADMIN_UI` | `enableAdminUi` | Whether or not to serve the administrative UI at `/admin` | `true` |
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
+| `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). Will enable OIDC if present. | Not set |
+| `LRSQL_OIDC_AUDIENCE` | `oidcAudience` | Optional OIDC audience for token claim verification. No effect if `LRSQL_OIDC_ISSUER` is not set. | Not set |
+| `LRSQL_OIDC_CLIENT_ID` | `oidcClientId` | An optional OIDC client ID for the SQL LRS Admin SPA. If provided, along with the `LRSQL_OIDC_ISSUER` and `LRSQL_OIDC_AUDIENCE` variables, will enable OIDC access to the Admin UI. | Not Set |
+| `LRSQL_OIDC_CLIENT_TEMPLATE` | `oidcClientTemplate` | An optional template to modify LRS Admin UI client OIDC configuration. | Not Set |
```
kelvinqian00, line 118: This should also say that this has no effect if `LRSQL_OIDC_ISSUER` is not set. Likewise if `LRSQL_OIDC_CLIENT_ID` is not set, for this particular variable.

milt: https://github.com/yetanalytics/lrsql/pull/206/commits/4fa3cc460aa1e896bd15d5efa9b1ffe3f2ced4d7

src/main/lrsql/ops/command/admin.clj
```diff
@@ -44,3 +45,31 @@
...
+               :tx transaction?
+               :input ads/ensure-admin-oidc-input-spec)
+  :ret ads/ensure-admin-ret-spec)
+
+(defn ensure-admin-oidc!
+  "Create a new admin with OIDC issuer or verify issuer of an existing admin.
+  Returns a map where `:result` is the account ID."
+  [bk tx {:keys [username oidc-issuer]
+          :as   ensure-input}]
+  (if-let [{extant-issuer :oidc_issuer ;; TODO: is not coerced to kebab?
```
kelvinqian00, line 64: It is indeed _not_ coerced to kebab! Notice in other query functions that you see the same thing (e.g. https://github.com/yetanalytics/lrsql/blob/4fa3cc460aa1e896bd15d5efa9b1ffe3f2ced4d7/src/main/lrsql/ops/query/document.clj#L30-35)

milt: Cool, I'll remove the comment

milt: https://github.com/yetanalytics/lrsql/pull/206/commits/a13f5f549c73962ae5214ee69e1e526dd6679a0a

doc/oidc.md
```diff
@@ -0,0 +1,88 @@
...
+### Identity Providers
+
+OIDC support is currently developed and tested against [Keycloak](https://www.keycloak.org/) but may work with other identity providers that implement the specification.
+
+#### Keycloak Demo
+
+This repository contains a Docker Compose file and configuration for a demo instance of keycloak that you can run locally to try out OIDC support. To run keycloak:
+
+    cd dev-resources/keycloak_demo
+    docker compose up
```
kelvinqian00, line 65: You could just put the make target here since it does the same thing.

And speaking of make targets, I also think you should add one that runs lrsql with these given OIDC config vars.

kelvinqian00:
As a final question, how feasible would it be to incorporate OIDC testing into CI? I could imagine that you can stand up a local Keycloack server using Testcontainers, set the OIDC config vars, then test SQL LRS as normal. Though I could easily be mistaken and it could be far more trouble than it is worse.

milt:
> As a final question, how feasible would it be to incorporate OIDC testing into CI?

Done some thinking about this. Probably not that _hard_, but a decent amount of work, as we use stuff from docker-compose that may or may not work well with TestContainers.

There is another way, which is similar to the tests in `poseidon`, where you seed the key cache or otherwise stub out the JWT issuance, allowing you to locally sign a jwt for tests. However I just don't see a huge value in integration testing OIDC in the first place.

kelvinqian00 (approved):

cliffcaseyyet (approved):
One very minor subjective thing. Also I pulled and ran this and noticed some insane lag time on performing an op (for instance making credentials), similar to what i saw in PS so i think somehow its my machine running KC that has all the lag, but please verify it doesnt take on the order of seconds for you. Also makes me wonder about the keycache because whenever i saw the lag on mine it was when keys first needed to be loaded, but here its happening all the time.

cliffcaseyyet (approved):

PR #207, opened 2022-02-21
[SQL-137] Admin API OIDC

milt:
[SQL-137][SQL-139] Enable OIDC auth + admin account shadowing.

This PR wraps the admin API with interceptors that will:
* Decode an OIDC token
* Check it for claim scope (right now there is just a blanket `admin` scope)
* Ensure that a local admin account is created to shadow
  * When an account exists, verifies that the OIDC issuer matches
* Passes the identity for use by identified admin operations, disabling the subsequent native JWT checks.

This enables OIDC clients to directly hit the admin API, enabling a wide range of use cases and not limiting administration to the provided UI.

Additional OIDC base improvements:
* Removed alternate config in favor of issuer
* Added optional token audience verification

[SQL-137]: https://yet.atlassian.net/browse/SQL-137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SQL-139]: https://yet.atlassian.net/browse/SQL-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -115,6 +115,7 @@ You may have noted that some options are not available:
 | `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
 | `LRSQL_OIDC_CONFIG` | `oidcConfig` | Custom OIDC configuration location | Not set |
 | `LRSQL_OIDC_AUDIENCE` | `oidcAudience` | Optional OIDC audience for token claim verification. | Not set |
+| `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER` (if provided). | `true` |
 | `LRSQL_JWKS_URI` | `jwksUri` | Custom JWKS [keyset](https://datatracker.ietf.org/doc/html/rfc7517#section-5) location | Not set |
 
 [<- Back to Index](index.md)
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -112,10 +112,8 @@ You may have noted that some options are not available:
 | `LRSQL_URL_PREFIX` | `urlPrefix` | The prefix of the webserver URL path, e.g. the prefix in `http://0.0.0.0:8080/xapi` is `/xapi`. Used when constructing the `more` value for multi-statement queries. | `/xapi` |
 | `LRSQL_ENABLE_ADMIN_UI` | `enableAdminUi` | Whether or not to serve the administrative UI at `/admin` | `true` |
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
-| `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
-| `LRSQL_OIDC_CONFIG` | `oidcConfig` | Custom OIDC configuration location | Not set |
+| `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). Will enable OIDC if present. | Not set |
 | `LRSQL_OIDC_AUDIENCE` | `oidcAudience` | Optional OIDC audience for token claim verification. | Not set |
 | `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER` (if provided). | `true` |
-| `LRSQL_JWKS_URI` | `jwksUri` | Custom JWKS [keyset](https://datatracker.ietf.org/doc/html/rfc7517#section-5) location | Not set |
 
 [<- Back to Index](index.md)
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -113,7 +113,7 @@ You may have noted that some options are not available:
 | `LRSQL_ENABLE_ADMIN_UI` | `enableAdminUi` | Whether or not to serve the administrative UI at `/admin` | `true` |
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
 | `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). Will enable OIDC if present. | Not set |
-| `LRSQL_OIDC_AUDIENCE` | `oidcAudience` | Optional OIDC audience for token claim verification. | Not set |
-| `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER` (if provided). | `true` |
+| `LRSQL_OIDC_AUDIENCE` | `oidcAudience` | Optional OIDC audience for token claim verification. No effect if `LRSQL_OIDC_ISSUER` is not set. | Not set |
+| `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER`. No effect if `LRSQL_OIDC_ISSUER` is not set. | `true` |
 
 [<- Back to Index](index.md)
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -114,6 +114,7 @@ You may have noted that some options are not available:
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
 | `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
 | `LRSQL_OIDC_CONFIG` | `oidcConfig` | Custom OIDC configuration location | Not set |
+| `LRSQL_OIDC_AUDIENCE` | `oidcAudience` | Optional OIDC audience for token claim verification. | Not set |
 | `LRSQL_JWKS_URI` | `jwksUri` | Custom JWKS [keyset](https://datatracker.ietf.org/doc/html/rfc7517#section-5) location | Not set |
 
 [<- Back to Index](index.md)
--- a/src/db/sqlite/lrsql/sqlite/record.clj
+++ b/src/db/sqlite/lrsql/sqlite/record.clj
@@ -179,6 +179,8 @@
   bp/AdminAccountBackend
   (-insert-admin-account! [_ tx input]
     (insert-admin-account! tx input))
+  (-insert-admin-account-oidc! [_ tx input]
+    (insert-admin-account-oidc! tx input))
   (-query-all-admin-accounts [_ tx]
     (query-all-accounts tx))
   (-delete-admin-account! [_ tx input]
--- a/src/db/sqlite/lrsql/sqlite/record.clj
+++ b/src/db/sqlite/lrsql/sqlite/record.clj
@@ -187,6 +187,8 @@
     (delete-admin-account! tx input))
   (-query-account [_ tx input]
     (query-account tx input))
+  (-query-account-oidc [_ tx input]
+    (query-account-oidc tx input))
   (-query-account-exists [_ tx input]
     (query-account-exists tx input))
 
--- a/src/db/sqlite/lrsql/sqlite/record.clj
+++ b/src/db/sqlite/lrsql/sqlite/record.clj
@@ -80,6 +80,8 @@
   (-update-all! [_ tx]
     (when (= 1 (:notnull (query-admin-account-passhash-notnull tx)))
       (update-schema-simple tx alter-admin-account-passhash-optional!))
+    (when-not (some? (query-admin-account-oidc-issuer-exists tx))
+      (alter-admin-account-add-openid-issuer! tx))
     (log/infof "sqlite schema_version: %d"
                (:schema_version (query-schema-version tx))))
 
--- a/src/db/sqlite/lrsql/sqlite/record.clj
+++ b/src/db/sqlite/lrsql/sqlite/record.clj
@@ -19,6 +19,7 @@
 
 ;; Schema Update Helpers
 
+#_{:clj-kondo/ignore [:unresolved-symbol]}
 (defn- update-schema-simple!
   "Given a tx and a db-fn that updates schema to remove CHECK or FOREIGN KEY or NOT NULL constraints or to add, remove, or change default values on a column, wraps it with an update to the schema version and associated checks.
 
--- a/src/db/sqlite/lrsql/sqlite/record.clj
+++ b/src/db/sqlite/lrsql/sqlite/record.clj
@@ -19,7 +19,7 @@
 
 ;; Schema Update Helpers
 
-(defn- update-schema-simple
+(defn- update-schema-simple!
   "Given a tx and a db-fn that updates schema to remove CHECK or FOREIGN KEY or NOT NULL constraints or to add, remove, or change default values on a column, wraps it with an update to the schema version and associated checks.
 
   See https://www.sqlite.org/lang_altertable.html Section 7 part 2"
@@ -79,7 +79,7 @@
     (create-credential-to-scope-table! tx))
   (-update-all! [_ tx]
     (when (= 1 (:notnull (query-admin-account-passhash-notnull tx)))
-      (update-schema-simple tx alter-admin-account-passhash-optional!))
+      (update-schema-simple! tx alter-admin-account-passhash-optional!))
     (when-not (some? (query-admin-account-oidc-issuer-exists tx))
       (alter-admin-account-add-openid-issuer! tx))
     (log/infof "sqlite schema_version: %d"
--- a/src/db/sqlite/lrsql/sqlite/record.clj
+++ b/src/db/sqlite/lrsql/sqlite/record.clj
@@ -1,5 +1,6 @@
 (ns lrsql.sqlite.record
-  (:require [com.stuartsierra.component :as cmp]
+  (:require [clojure.tools.logging :as log]
+            [com.stuartsierra.component :as cmp]
             [lrsql.backend.protocol :as bp]
             [lrsql.backend.data :as bd]
             [lrsql.init :refer [init-hugsql-adapter!]]
@@ -61,7 +62,16 @@
     (create-credential-table! tx)
     (create-credential-to-scope-table! tx))
   (-update-all! [_ tx]
-    (alter-admin-account-passhash-optional! tx))
+    (log/info "Updating DB schema...")
+    (log/infof "schema version %s" (query-schema-version tx))
+    (when (= 1 (:notnull (query-admin-account-passhash-notnull tx)))
+      (let [{schema-version :schema_version} (query-schema-version tx)]
+        (log/info "Setting admin account passhash to optional")
+        (enable-writable-schema-snip tx)
+        (alter-admin-account-passhash-optional! tx)
+        (update-schema-version! tx {:schema-version (inc schema-version)})
+        (disable-writable-schema-snip tx)
+        (integrity-check-snip tx))))
 
   bp/BackendUtil
   (-txn-retry? [_ _ex]
--- a/src/db/sqlite/lrsql/sqlite/record.clj
+++ b/src/db/sqlite/lrsql/sqlite/record.clj
@@ -60,9 +60,8 @@
     (create-admin-account-table! tx)
     (create-credential-table! tx)
     (create-credential-to-scope-table! tx))
-  (-update-all! [_ _]
-    ;; No-op for now; add functions if updates are needed
-    nil)
+  (-update-all! [_ tx]
+    (alter-admin-account-passhash-optional! tx))
 
   bp/BackendUtil
   (-txn-retry? [_ _ex]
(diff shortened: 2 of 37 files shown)


src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -61,7 +62,16 @@
...
     (create-credential-to-scope-table! tx))
   (-update-all! [_ tx]
-    (alter-admin-account-passhash-optional! tx))
+    (log/info "Updating DB schema...")
+    (log/infof "schema version %s" (query-schema-version tx))
+    (when (= 1 (:notnull (query-admin-account-passhash-notnull tx)))
+      (let [{schema-version :schema_version} (query-schema-version tx)]
+        (log/info "Setting admin account passhash to optional")
+        (enable-writable-schema-snip tx)
+        (alter-admin-account-passhash-optional! tx)
```
milt, line 71: This attempted schema edit fails, the previous `enable-writable-schema-snip` does not appear to apply:

```
Execution error (SQLiteException) at org.sqlite.core.DB/newSQLException (DB.java:1012).
[SQLITE_ERROR] SQL error or missing database (table sqlite_master may not be modified)
```

src/main/lrsql/system/lrs.clj
```diff
@@ -261,6 +261,12 @@
           input (admin-input/delete-admin-input account-id)]
       (jdbc/with-transaction [tx conn]
         (admin-cmd/delete-admin! backend tx input))))
+  (-ensure-account-oidc
+    [this username oidc-issuer]
+    (let [conn  (lrs-conn this)]
+      (jdbc/with-transaction [tx conn]
+        (admin-cmd/ensure-admin-oidc! backend tx {:username    username
+                                                  :oidc-issuer oidc-issuer}))))
```
kelvinqian00, line 269: Now that I have the bigger picture, I think you should call an `ensure-admin-oidc-input` function here, then generate the primary key if need be at the op level.

kelvinqian00:
Looking through this, I'm still confused as to where and why OIDC needs to be ensured upon every _xAPI_ authorization (as opposed to each admin account authorization). It's not obvious to me looking at the resource routes or the protocols.

src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -16,6 +17,22 @@
 (u/def-hugsql-db-fns "lrsql/sqlite/sql/update.sql")
 (u/def-hugsql-db-fns "lrsql/sqlite/sql/delete.sql")
 
+;; Schema Update Helpers
+
+(defn- update-schema-simple
```
kelvinqian00, line 22: Should have the `!` at the end of this mutating fn.

src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -16,6 +17,22 @@
...
+  See https://www.sqlite.org/lang_altertable.html Section 7 part 2"
+  [tx db-fn]
+  (enable-writable-schema! tx)
+  (db-fn tx)
+  (update-schema-version!
+   tx
+   (update (query-schema-version tx)
+           :schema_version inc))
+  (disable-writable-schema! tx)
+  (run-integrity-check tx))
```
kelvinqian00, line 34: I wonder if there should be something more sophisticated here if `run-integrity-check` fails, like a rollback mechanism.

kelvinqian00:
I just added @cliffcaseyyet as a reviewer, which was probably something you were going to do once he got back. But I would specifically like him to look at the SQLite changes to see whether or not there are any better ways to perform migrations on that DBMS.

src/main/lrsql/system/lrs.clj
milt: Yep, that sounds solid for ensure. I think I see what you're saying about `insert-admin!` too, because it fires that query first.

src/db/sqlite/lrsql/sqlite/record.clj
milt: indeed

src/db/sqlite/lrsql/sqlite/record.clj
milt: I'm a little suspicious of `PRAGMA integrity_check` because it passed on a previous version where I was actually using incorrect schema for this kind of update (it had line breaks, which are bad apparently). It didn't fail until the _next time_ I tried to update the table. Rollback is challenging, but what you describe is done in other systems. For instance in Rails I'm pretty sure the procedure was that for every migration you would define an "up" and a "down" procedure. Needs more research given how limited the schema functionality in sqlite is.

milt:
> Looking through this, I'm still confused as to where and why OIDC needs to be ensured upon every xAPI authorization (as opposed to each admin account authorization). It's not obvious to me looking at the resource routes or the protocols.

Yep, this is one of the harder things to grok about this, your confusion is warranted. Basically, consider two scenarios:
1. OIDC is used to authenticate an admin, giving them a "native" app token to do subsequent operations. Ensuring OIDC shadow accounts only happens at the authentication route. This is effectively token exchange, you are exchanging your OIDC token for an LRSQL token and then doing stuff with it.
2. OIDC is accepted on all admin routes directly. Ensuring OIDC accounts happens any time an OIDC token is used. With this kind of direct federation, all you need is an OIDC token.

In scenario 1, a client MUST use the auth route first to exchange that token. That's an extra step, but that in and of itself isn't too big a deal (though annoying). The complexity comes in _over time_: lrsql doesn't really have any conception of "refresh tokens" or anything like that, so when the lrsql token expires, you need to go back and do it all over again. This means that in the meantime, in addition to keeping the OIDC token refreshed you need to maintain a lifecycle for the lrsql token, something completely outside of OIDC's purview.

Contrast that with scenario 2, where there is only ever one OIDC token involved. There is no extra request, and all policy (such as exp times) and behavior comes from the OIDC provider, not lrsql. The OIDC client automatically refreshes and handles lifecycle. For a system integrator coming in this is much more what they would expect from OIDC support, the extra request and lifecycle are unknown to that spec. Therefore software that is already OIDC-enabled will "just work" without custom calls and state. While neither scenario is "stateless" per se, scenario 2 has less to worry about.

src/main/lrsql/admin/interceptors/oidc.clj
```diff
@@ -0,0 +1,90 @@
...
+                   {:status 401
+                    :body   {:error "OIDC Issuer Mismatch!"}})
+            (-> ctx
+                disable-jwt-interceptors
+                (assoc-in [::jwt/data :account-id] result)
+                ;; NOTE: I do this to follow what is done for native JWT
+                ;; Don't think this app actually uses a session at all?
+                (assoc-in
+                 [:request :session ::jwt/data :account-id]
+                 result))))
```
milt, line 89: Was curious about this, as I found several places where session values are set, but none where they get used. Additionally no kind of session config or store is present in the service map. When I see it in the jwt interceptors, is it vestigial?

src/main/lrsql/util/auth.clj
```diff
@@ -14,7 +14,8 @@
   {"all"                  :scope/all
    "all/read"             :scope/all.read
    "statements/read"      :scope/statements.read
-   "statements/write"     :scope/statements.write})
+   "statements/write"     :scope/statements.write
+   "admin"                :scope/admin})
```
milt, line 18: I'm calling this out as a possible smell, I added the `admin` scope (expressing the ability to do admin things) to the same scopes used for xAPI. This was more out of a desire to dedupe code than anything else, but it might not be great.

src/db/sqlite/lrsql/sqlite/record.clj
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/45d4791186a35b00ce79a0b0c1306cf8edb4ca72

src/main/lrsql/system/lrs.clj
milt: Done for ensure on https://github.com/yetanalytics/lrsql/pull/207/commits/1e2526a4fe40d895f9a925606b31d48075510573, added the `add-primary-key` fn to util for the purpose. I'll go ahead and try a similar thing for the vanilla

milt: for the vanilla admin https://github.com/yetanalytics/lrsql/pull/207/commits/785148c4818620f5c3e6fd817b4040405cec4c49

kelvinqian00:
> Yep, this is one of the harder things to grok about this...
I think a lot of my confusion comes from when you say all xAPI authentication, I imagined you would need OIDC when _performing LRS operations_, like inserting and querying statements. From what you said and from what I saw in the code, it seems like you actually meant "all _admin_ ops, not just logging in."

src/main/lrsql/admin/interceptors/oidc.clj
kelvinqian00: I believe I added that code so that the account-id data is in two places (the JWT as well as higher up in the Pedestal context) as a form of redundancy. I guess it's not strictly needed, but I think it's good to have that extra info in the JWT.

src/main/lrsql/util/auth.clj
kelvinqian00: These auth levels are strictly for statement insert/query and other LRS ops. I don't think it makes sense to add an "admin" scope since in lrsql _every_ user is effectively an admin, i.e. they can create or delete both accounts and API key pairs.

src/test/lrsql/util/auth_test.clj
```diff
@@ -119,6 +119,14 @@
              :scopes         #{}}
       false {:request-method :delete
              :path-info      "xapi/activities/state"
+             :scopes         #{}}
+      ;; Admin Scope
+      ;; Not used for xAPI, only for Admin requests
+      true {:request-method :get
+            :path-info      "/admin/account"
```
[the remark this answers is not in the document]
kelvinqian00: Huh I guess you're right. We'll it's fortunate that the function isn't particularly strict about the nature of the path info string (e.g. it only checks if `statements` is a suffix for certain scopes). I guess you could make it more strict, but there's a small possibility it could be backwards incompatible.

milt:
> I think a lot of my confusion comes from when you say "all xAPI authentication", I imagined you would need OIDC when performing LRS operations, like inserting and querying statements. From what you said and from what I saw in the code, it seems like you actually meant "all admin ops, not just logging in."

Both are true...
* You can (but are not required to) use an OIDC token to authenticate for LRS actions like posting statements. This covers the (somewhat narrow) use case of an OIDC-enabled web app sending statements to the LRS. Note that this was done in a separate PR, already merged to the feature branch.
* You can also use OIDC, with the additional `admin` scope, to access all admin API actions.
  * You _could_ use it to just log in and get an LRSQL token, a la scenario 1
  * Or you could use the OIDC token directly, for instance to create a set of long-lived basic auth credentials

src/main/lrsql/admin/interceptors/oidc.clj
milt: Sure, I have no problem with it being there, it just reads like there is going to be some session to hold that data (per ring conventions). That said, you will often see duplicate data placed on the context _and_ the request, this is generally to provide it to function-style terminal interceptors ("handlers"), that only ever get the request.

src/main/lrsql/util/auth.clj
milt: > These auth levels are strictly for statement insert/query and other LRS ops.

Agreed, it doesn't belong there

> I don't think it makes sense to add an "admin" scope since in lrsql every user is effectively an admin, i.e. they can create or delete both accounts and API key pairs.

I don't agree with this. If we don't require *any* scope for admin actions, there is nothing to differentiate a user who for instance has the `statements/write` scope from one that can create/modify accounts and credentials. Proper OIDC support for admin stuff needs a scope, I just don't think it should be here per se.

milt: Or put another way, it makes the xAPI scopes meaningless if any identity on the IDP can just make some new creds with the `all` scope just by virtue of being in the system

milt: Consider an organization with an existing IDP. They might have 1000 users in the IDP. Within that, they want 500 of those users to be able to post statements and only 10 to be able to administer it. Scopes!

kelvinqian00: I see - before we had a clear distinction of who's an admin and who's just a LRS statement reader/writer based on what they use to authenticate (API key pairs in the latter case). If you're using OIDC for both, that distinction breaks down, so the old auth model of having just scopes for statements becomes obsolete.

milt: Yep, that's pretty much it. You have to have some way to distinguish them, it's implicit in the local/vanilla auth (and that's fine), but you need to differentiate if you're federating with a larger system

kelvinqian00: In that case there's definitely going to be some surgery done to the `authorize-action` function. I wonder if any additional changes should be made to the `lrs` library to make it easier to work with OIDC (I'm certain it's not required though).

milt: Well again, the LRS lib knows nothing about administration, and I don't know that it should. It delegates this stuff downstream because it is hard to generalize. See https://github.com/yetanalytics/lrsql/pull/207/commits/29126d851ccf3109e572963561bf2b46983650da for an extraction of the admin scope to the OIDC ns

src/main/lrsql/util/oidc.clj
```diff
@@ -118,11 +151,28 @@
...
+(defn authorize-admin-action
+  "Given a pedestal context and an OIDC admin auth identity, authorize or deny."
+  [{{:keys [request-method path-info]} :request
+    :as _ctx}
+   {:keys [scopes]
+    :as _auth-identity}]
+  {:result
+   (boolean
+    (and (cs/starts-with? path-info "/admin")
+         (contains? scopes :scope/admin)))})
```
kelvinqian00, line 178: So you're basing the difference on the REST endpoint path, not anything fundamental about the OIDC token itself? (I'm assuming it's because the latter is impossible and that an OIDC token is an OIDC token.)

milt: Well `scope` is itself a claim on OIDC access tokens, for this exact case (the LRS/admin routes are a "resouce server"). So prior to this the scope claim is decoded and matched to known scopes, subject to the custom user prefix. I just used `:path-info` because we only have the one scope, in a more complex scenario it would probably use more than just that.

milt: For reference, when my OIDC client (which is just a modified version of the re-oidc demo rn) makes its initial auth request to the IDP, it specifies that it wants `all`, `admin`, etc. scopes, and that its "audience" includes the uri (known to the IDP) of the LRS. Then the IDP can (however it wants) decide to grant these or not. If it does, they are included in the token claims (and thus verifiable).

src/db/sqlite/lrsql/sqlite/record.clj
cliffcaseyyet: Yeah we probably do not want to get into an automated rollback procedure here. That's very very stricky stuff. Recommendation in enterprise should be snapshot before upgrade, restore and version reduction for rollback.

src/main/lrsql/system/lrs.clj
milt: ended up reverting the changes to vanilla admin https://github.com/yetanalytics/lrsql/commit/785148c4818620f5c3e6fd817b4040405cec4c49, ran into weird spec problems that are out of scope for this anyhow. Honestly the chance of generating an unneeded ID is much lower with that than when ensuring OIDC admins.

kelvinqian00: To quote someone wise, "well we tried."

src/db/sqlite/lrsql/sqlite/record.clj
kelvinqian00: Do you recommend that such a snapshotting mechanism be implemented here?

doc/env_vars.md
```diff
@@ -112,8 +112,8 @@ You may have noted that some options are not available:
...
 | `LRSQL_ENABLE_STMT_HTML` | `enableStmtHtml` | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature. | `true` |
-| `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) | Not set |
-| `LRSQL_OIDC_CONFIG` | `oidcConfig` | Custom OIDC configuration location | Not set |
-| `LRSQL_JWKS_URI` | `jwksUri` | Custom JWKS [keyset](https://datatracker.ietf.org/doc/html/rfc7517#section-5) location | Not set |
+| `LRSQL_OIDC_ISSUER` | `oidcIssuer` | OIDC Issuer address used for [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). Will enable OIDC if present. | Not set |
+| `LRSQL_OIDC_AUDIENCE` | `oidcAudience` | Optional OIDC audience for token claim verification. | Not set |
+| `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER` (if provided). | `true` |
```
kelvinqian00, line 117: Wait I thought you were just going to keep `LRSQL_OIDC_ISSUER`, i.e. only have one OIDC-related env var.

src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -16,6 +17,22 @@
 (u/def-hugsql-db-fns "lrsql/sqlite/sql/update.sql")
 (u/def-hugsql-db-fns "lrsql/sqlite/sql/delete.sql")
 
+;; Schema Update Helpers
+
+(defn- update-schema-simple!
```
kelvinqian00, line 22: You'd want to add a `#_{:clj-kondo/ignore [:unresolved-symbol]}` here since I'm getting errors in VSCode. (Of course double check that they're actually defined and usable before you do so.)

doc/env_vars.md
milt: The vars I removed were alternates for `LRSQL_OIDC_ISSUER` (`LRSQL_OIDC_CONFIG` and `LRSQL_JWKS_URI`). If you just provide issuer, you can use the LRS as an OIDC resource server (for xAPI/admin api). If you provide `LRSQL_OIDC_AUDIENCE`, it will additionally verify token audience (this will be recommended in the docs, which are pending). `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` toggles an additional sanity check that the issuer you put in matches the one in the remote OIDC config.

src/main/lrsql/ops/command/admin.clj
```diff
@@ -44,3 +45,32 @@
...
+            id            :id} (bp/-query-account-oidc
+                                bk tx {:username username})]
+    {:result
+     (if (= oidc-issuer extant-issuer)
+       id
+       :lrsql.admin/oidc-issuer-mismatch-error)}
+    (let [{:keys [primary-key]
+           :as   insert-input} (admin-i/insert-admin-oidc-input
+                                ensure-input)]
+      (do
```
kelvinqian00, line 74: This `do` is unneeded.

src/main/lrsql/init/oidc.clj
```diff
@@ -84,6 +115,21 @@
...
 
+(s/fdef admin-interceptors
+  :args (s/cat :config partial-config-spec)
+  :ret (s/every i/interceptor?))
+
+(defn admin-interceptors
+  "Given a webserver config, return a (possibly empty) vector of interceptors
+  for use with the admin API. These validate token claims and ensure an admin
+  account is made."
+  [{:keys [oidc-issuer] :as config}]
```
kelvinqian00, line 126: This `config` var results in an unused binding warning.

src/main/lrsql/util/oidc.clj
```diff
@@ -88,3 +112,67 @@
...
+  "For the given context, return a valid OIDC admin auth identity from token
+  claims.
+  args:
+    ctx - Pedestal context that may contain claims.
+    scope-prefix - Prefix to add to expected scopes."
+  [ctx
+   scope-prefix]
+  (when-let [token (:com.yetanalytics.pedestal-oidc/token ctx)]
+    (let [{:keys [scope iss sub]
+           :as   claims} (get-in ctx
```
kelvinqian00, line 147: `token` and `claims` are both unused bindings.

src/main/lrsql/util/oidc.clj
```diff
@@ -88,3 +112,67 @@
...
+        ::unauthorized))))
+
+(s/fdef authorize-admin-action
+  :args (s/cat :ctx           (s/keys :req-un [::auth/request])
+               :auth-identity (s/keys :req-un [::scopes]))
+  :ret (s/keys :req-un [::auth/result]))
+
+(defn authorize-admin-action
+  "Given a pedestal context and an OIDC admin auth identity, authorize or deny."
+  [{{:keys [request-method path-info]} :request
```
kelvinqian00, line 171: `request-method` is an unused binding.

src/db/sqlite/lrsql/sqlite/record.clj
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/4fc65748134088033d169f93230f1c513b26c5ec

doc/env_vars.md
kelvinqian00: Maybe also add in the `LRSQL_OIDC_AUDIENCE` and `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` descriptions that they will be unused if `LRSQL_OIDC_ISSUER` is not set.

src/main/lrsql/ops/command/admin.clj
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/0d2cfe97791fa990b600eaef06e9ac19ab7dfcab

doc/env_vars.md
milt: thanks, I meant to do that yesterday

src/main/lrsql/init/oidc.clj
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/d2db9b892d0a5df1e39c6d2a7609f76d52f79493

src/main/lrsql/util/oidc.clj
```diff
@@ -5,47 +5,71 @@
             [clojure.tools.logging :as log]
             [com.yetanalytics.lrs.auth :as lrs-auth]
             [lrsql.init.authority :as authority]
+            [lrsql.spec.admin :as admin]
             [lrsql.spec.config :as config]
             [lrsql.util.auth :as auth])
   (:import [java.io File]))
```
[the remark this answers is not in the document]
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/d2db9b892d0a5df1e39c6d2a7609f76d52f79493

src/main/lrsql/util/oidc.clj
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/d2db9b892d0a5df1e39c6d2a7609f76d52f79493

src/main/lrsql/util/oidc.clj
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/d2db9b892d0a5df1e39c6d2a7609f76d52f79493

kelvinqian00 (approved):

doc/env_vars.md
milt: https://github.com/yetanalytics/lrsql/pull/207/commits/e8c90be1e5b0218a5aa3bf6df72cc44230c3a622

src/db/sqlite/lrsql/sqlite/record.clj
cliffcaseyyet: hard no

cliffcaseyyet (approved):
dig it

PR #211, opened 2022-03-07
SQL-140 Keycloak Demo

milt:
Set up a keycloak demo for ongoing OIDC development.

To create and start the Keycloak demo environment on port 8081:

    make keycloak-demo

To start an ephemeral instance of SQL LRS configured to use it for OIDC authentication:

    make ephemeral-oidc

The [`lrsql_testing`](https://github.com/yetanalytics/re-oidc/tree/lrsql_testing) branch of `re-oidc` can be used to verify that authentication and LRS communication is working.


Makefile
```diff
@@ -51,6 +51,10 @@ ephemeral: resources/public/admin
 ephemeral-prod: resources/public/admin
 	clojure -M:db-h2 -m lrsql.h2.main --persistent false
```
kelvinqian00, line 52: Do we still need this `ephemeral-prod` target now that we have `ephemeral-oidc`?

src/db/h2/lrsql/h2/main.clj
```diff
@@ -10,8 +10,10 @@
   "Run an H2-backed LRSQL instance based on the `:test-h2` (if `:persistent?`
    is set to `true`) or `:test-h2-mem` (if not) config profile. For use with
    `clojure -X:db-h2`."
-  [{:keys [persistent?]}]
-  (let [profile (if persistent? :test-h2 :test-h2-mem)]
+  [{:keys [persistent?
+           override-profile]}]
+  (let [profile (or override-profile
+                  (if persistent? :test-h2 :test-h2-mem))]
     (component/start (system/system h2-backend profile))))
```
kelvinqian00, line 17: Maybe you should just have a single `:profile` argument if it's no longer a binary choice.

milt: Yeah, the way I did it was specifically to not change the interface as it exists.

Makefile
kelvinqian00: Makes sense - to be clear does that mean `ephemeral-prod` is useful outside of keycloak?

milt: Yes, I believe so. Specifically thinking here about cases where we're following up on a submitted issue or doing paid support.

kelvinqian00:
I also noticed that you added a `keycloak` dir at the top level. Is there a reason for that instead of locating that dir within, say, `dev-resources`?

milt:
It's by convention, matches our other projects. I don't mind moving it, should work just as well.

kelvinqian00 (approved):

PR #217, opened 2022-03-23
SQL-144 OIDC local admin toggle

milt:
[SQL-144] When started with settings to provide admin OIDC access, the `LRSQL_OIDC_ENABLE_LOCAL_ADMIN` flag determines whether or not local admin accounts will still be allowed to do things.

[SQL-144]: https://yet.atlassian.net/browse/SQL-144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -117,5 +117,6 @@ You may have noted that some options are not available:
 | `LRSQL_OIDC_CLIENT_ID` | `oidcClientId` | An optional OIDC client ID for the SQL LRS Admin SPA. If provided, along with the `LRSQL_OIDC_ISSUER` and `LRSQL_OIDC_AUDIENCE` variables, will enable OIDC access to the Admin UI. | Not Set |
 | `LRSQL_OIDC_CLIENT_TEMPLATE` | `oidcClientTemplate` | An optional template to modify LRS Admin UI client OIDC configuration. | Not Set |
 | `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER`. No effect if `LRSQL_OIDC_ISSUER` is not set. | `true` |
+| `LRSQL_OIDC_ENABLE_LOCAL_ADMIN` | `oidcEnableLocalAdmin` | Whether or not to enable local administrative account login when `LRSQL_OIDC_CLIENT_ID` is set. | `false` |
 
 [<- Back to Index](index.md)
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -117,6 +117,6 @@ You may have noted that some options are not available:
 | `LRSQL_OIDC_CLIENT_ID` | `oidcClientId` | An optional OIDC client ID for the SQL LRS Admin SPA. If provided, along with the `LRSQL_OIDC_ISSUER` and `LRSQL_OIDC_AUDIENCE` variables, will enable OIDC access to the Admin UI. | Not Set |
 | `LRSQL_OIDC_CLIENT_TEMPLATE` | `oidcClientTemplate` | An optional template to modify LRS Admin UI client OIDC configuration. | Not Set |
 | `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER`. No effect if `LRSQL_OIDC_ISSUER` is not set. | `true` |
-| `LRSQL_OIDC_ENABLE_LOCAL_ADMIN` | `oidcEnableLocalAdmin` | Whether or not to enable local administrative account authentication and login when `LRSQL_OIDC_ISSUER` is set. | `false` |
+| `LRSQL_OIDC_ENABLE_LOCAL_ADMIN` | `oidcEnableLocalAdmin` | Whether or not to enable local administrative account authentication, login and management when `LRSQL_OIDC_ISSUER` is set. | `false` |
 
 [<- Back to Index](index.md)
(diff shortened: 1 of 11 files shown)


src/main/lrsql/init/oidc.clj
```diff
@@ -250,14 +256,17 @@
   "Given webserver and LRS configs, return a vector of interceptors to apply to
   Admin UI routes. If webserver oidc-client-id is not specified, returns an
   empty vector."
-  [{:keys [oidc-issuer
-           oidc-client-id] :as webserver-config}
+  [{:keys                           [oidc-issuer
+           oidc-client-id
+           oidc-enable-local-admin] :as webserver-config}
```
kelvinqian00, line 261: Formatting boo-boo?

resources/lrsql/config/prod/default/webserver.edn
```diff
@@ -18,4 +18,5 @@
  :oidc-audience             #env LRSQL_OIDC_AUDIENCE
  :oidc-client-id            #env LRSQL_OIDC_CLIENT_ID
  :oidc-client-template      #or [#env LRSQL_OIDC_CLIENT_TEMPLATE "config/oidc_client.json.template"]
- :oidc-verify-remote-issuer #or [#env LRSQL_OIDC_VERIFY_REMOTE_ISSUER true]}
+ :oidc-verify-remote-issuer #or [#env LRSQL_OIDC_VERIFY_REMOTE_ISSUER true]
+ :oidc-enable-local-admin   #boolean #or [#env LRSQL_OIDC_ENABLE_LOCAL_ADMIN false]}
```
kelvinqian00, line 22: Not really related to this PR but shouldn't `:oidc-verify-remote-issuer` also have a `#boolean` tag?

doc/env_vars.md
```diff
@@ -117,5 +117,6 @@ You may have noted that some options are not available:
 | `LRSQL_OIDC_CLIENT_ID` | `oidcClientId` | An optional OIDC client ID for the SQL LRS Admin SPA. If provided, along with the `LRSQL_OIDC_ISSUER` and `LRSQL_OIDC_AUDIENCE` variables, will enable OIDC access to the Admin UI. | Not Set |
 | `LRSQL_OIDC_CLIENT_TEMPLATE` | `oidcClientTemplate` | An optional template to modify LRS Admin UI client OIDC configuration. | Not Set |
 | `LRSQL_OIDC_VERIFY_REMOTE_ISSUER` | `oidcVerifyRemoteIssuer` | Verify on startup that the issuer in remote configuration matches `LRSQL_OIDC_ISSUER`. No effect if `LRSQL_OIDC_ISSUER` is not set. | `true` |
+| `LRSQL_OIDC_ENABLE_LOCAL_ADMIN` | `oidcEnableLocalAdmin` | Whether or not to enable local administrative account login when `LRSQL_OIDC_CLIENT_ID` is set. | `false` |
```
kelvinqian00, line 120: Since this var only takes effect if `LRSQL_OIDC_CLIENT_ID` is set, that means this wouldn't have an effect if the user is logging in via OIDC _without_ using the admin UI (e.g. they're CURL-ing into the LRS). How important would taking into account this edge case be?

src/main/lrsql/init/oidc.clj
milt: Yeah, I use `clojure-align` on emacs and sometimes it gets a bit spicy, thanks

milt: https://github.com/yetanalytics/lrsql/pull/217/commits/8f9de18a3ee34d574f4c30c33dca01020d730e1e

doc/env_vars.md
milt: Super important! I actually brought this up on Slack this morning before reading this, you're spot on. There are a couple things that need coverage/thought here:
* This doc is actually incorrect, because setting this to false will actually prevent any local admin login, even if an `oidc-client-id` is not present. I'll update that.
  * This makes the admin FE effectively useless if `oidc-client-id` is not also provided, but a non-functional login screen is fine
* The local admin account CRUD still works for direct OIDC token users when `oidc-enable-local-admin` is false, this is not desirable.

milt: doc update https://github.com/yetanalytics/lrsql/pull/217/commits/dbf89319e5a84e7b6ee053fe532fe5a5c6c00d29

resources/lrsql/config/prod/default/webserver.edn
milt: It should!

doc/env_vars.md
milt: disable account routes: https://github.com/yetanalytics/lrsql/pull/217/commits/fa1ffbbaa996d6f9191aa068881f487793ac7e68

resources/lrsql/config/prod/default/webserver.edn
milt: https://github.com/yetanalytics/lrsql/pull/217/commits/55befad2a048fb541a42b00ef2efb210cdede65d

src/main/lrsql/init/oidc.clj
```diff
@@ -128,11 +130,15 @@
   "Given a webserver config, return a (possibly empty) vector of interceptors
   for use with the admin API. These validate token claims and ensure an admin
   account is made."
-  [{:keys [oidc-issuer]}]
+  [{:keys [oidc-issuer
+           oidc-enable-local-admin]}]
```
kelvinqian00, line 134: I notice that you use `oidc-enable-local-admin` in multiple places in the code, in multiple namespaces. I feel like this is fairly brittle, since you would have multiple places where you could introduce bugs relating to this config variable. I'm personally not sure what's the best way to refactor this, but it's something to consider.

milt: I'm not sure what the problem is? It is a config that is needed in several places, so I use it.

kelvinqian00: Sure, but if you could reduce the number of places it's needed in, I think it would be great (even if it's impossible to only use it in a single place).

milt: Sometimes complex features are just complex, yeah.

kelvinqian00 (approved):

PR #222, opened 2022-04-25
SQL-147 Add suppression for core.async nvd FP

milt:
Per https://github.com/jeremylong/DependencyCheck/issues/4384, we have a false positive on nvd scan. Add suppression file to clear core.async, at least until it is resolved.

Note that this should follow the merge + new version resulting from https://github.com/yetanalytics/actions/pull/9


Makefile
```diff
@@ -76,7 +76,7 @@ bench-async:
 # Vulnerability check
 
 target/nvd:
-	clojure -Xnvd check :classpath '"'"$$(clojure -Spath -A:db-h2:db-sqlite:db-postgres)"'"'
+	clojure -Xnvd check :classpath '"'"$$(clojure -Spath -A:db-h2:db-sqlite:db-postgres)"'"' :config-filename '"nvd_config.json"'
```
kelvinqian00, line 79: Feel like `nvd_config.json` (and `nvd_suppression.xml`) could go in the `dev_resources` folder.

kelvinqian00: (Unless the `:nvd` alias requires that the files be at the top level)

milt: sure

.nvd/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "nvd": {"suppression-file": ".nvd/suppression.xml"}
```
milt, line 2: @kelvinqian00 this is a good example of why we put config in the root, file paths in config files. This works (because the .nvd/ added), but is immediately more confusing because it knows nothing of the relative path in the dir. In this case everything works so let's just go with it.

Makefile
milt: https://github.com/yetanalytics/lrsql/pull/222/commits/2b7bf49f3c6c5af6cd7ab236a851cbdf4dc1475f

kelvinqian00 (approved):

PR #234, opened 2022-06-20
SQL-150 Overwrite Statement Authority

milt:
[SQL-150] SQL LRS currently does not overwrite authority with the operator-defined authority, a bug. This PR forces overwrite of all incoming statement authority objects.

[SQL-150]: https://yet.atlassian.net/browse/SQL-150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/authority.md
+++ b/doc/authority.md
@@ -2,7 +2,7 @@
 
 # Authority Configuration
 
-The SQL LRS allows for configuration of the Authority included in xAPI Statements that are written to the LRS.
+The SQL LRS allows for configuration of the Authority included in xAPI statements that are written to the LRS. This authority will overwrite any authority present on incoming statements.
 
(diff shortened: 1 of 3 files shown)


kelvinqian00:
Two things left:
- [ ] Either delete the `NOTE` comment above `prepare-statements` or revise it to say the opposite (i.e. that authority will be overwritten since we _cannot_ assume a strong trust relationship between the LRS and the LRP).
- [ ] Add to the documentation that the authority will be overwritten. (It might be that any LRS user would assume this anyways, but it is always nice to clarify this.)

kelvinqian00 (approved):

Issue #237, opened 2022-07-08
Changing the log level


catmeme:
What is the recommended way to change the log level?

milt:
@catmeme that's a really good question! Right now it doesn't look like there's an easy way to do it at runtime, as we include the logback.xml in the jar. I'll confer with the team, but it seems to me there's two ways it could work (that are not mutually exclusive):

* Add a classpath arg to our run scripts that points to a logback file outside the jar, so users can modify
* Add an env var to override base logging level via the logback API

I'll do some investigating and get back to you.

catmeme:
Thanks for this!

PR #238, opened 2022-07-08
SQL-154 allow runtime log level control with env var

milt:
[SQL-154] Per https://github.com/yetanalytics/lrsql/issues/237, provide a simple way for users to set the runtime log level.

[SQL-154]: https://yet.atlassian.net/browse/SQL-154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/lrsql/system.clj
+++ b/src/main/lrsql/system.clj
@@ -1,6 +1,7 @@
 (ns lrsql.system
   (:require [com.stuartsierra.component :as component]
             [lrsql.system.database :as db]
+            [lrsql.system.logger :as logger]
             [lrsql.system.lrs :as lrs]
             [lrsql.system.webserver :as webserver]
             [lrsql.init.config :refer [read-config]]))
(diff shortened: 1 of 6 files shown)


src/main/lrsql/system.clj
```diff
@@ -13,18 +14,20 @@
...
-                      [:backend])
+                      [:logger :backend])
          :lrs        (component/using
                       (lrs/map->LearningRecordStore {})
-                      [:connection :backend])
+                      [:logger :connection :backend])
          :webserver  (component/using
                       (webserver/map->Webserver {})
-                      [:lrs]))
+                      [:logger :lrs]))
```
kelvinqian00, line 30: Didn't realize that you could have dependent components that _aren't_ record fields.

I'm not sure if you really need all these `using ... :logger` calls on the components above `:backend`, since they all build on top of each other.

milt: > Didn't realize that you could have dependent components that aren't record fields.

Yep, it's also an associative map. It's weird.

> I'm not sure if you really need all these using ... :logger calls on the components above :backend, since they all build on top of each other.

I don't love it, but since we want it to happen _before_ anything that could log at init in the other components, this seems to be the best way.

milt: Oh I see what you're saying, because everything else depends on backend already. Yeah, I can save some code that way I guess

milt: https://github.com/yetanalytics/lrsql/pull/238/commits/a5daeeaaf185be7b7169dd47f460e0cc302bfb8b

kelvinqian00 (approved):

Issue #235, opened 2022-06-27
Enable socket unix for postgres


florianrusso:
Hi,
we run SQLRS on cloud run, in order to be secure we need java packages to handle unix sockets.
What do you think about this? I will be more than happy to open a new pull request but I’m not sure how to do that not being used to clojure.
I added the dependencies
```edn
  com.kohlschutter.junixsocket/junixsocket-core  {:mvn/version “2.5.0”}
```
```
Downloading: com/kohlschutter/junixsocket/junixsocket-native-common/2.5.0/junixsocket-native-common-2.5.0.jar from central
Error building classpath. Could not find artifact com.kohlschutter.junixsocket:junixsocket-core:jar:2.5.0 in central (https://repo1.maven.org/maven2/)
make: *** [test-h2] Error 1

```
I didn’t manage to make tests work. due to unfound jar, so I tried
```edn
  com.kohlschutter.junixsocket/junixsocket-core  {:mvn/version “2.3.4"}
```
```
clojure -M:test -m lrsql.test-runner --database h2
Syntax error compiling at (lrsql/util/cert.clj:50:39).
Unable to find static field: md5WithRSAEncryption_oid in class sun.security.x509.AlgorithmId
```
seems to work during install, but I have an issue to launch the tests locally (with or without any change) and thought it would be a better idea to use docker instead.

to test I updated the docker compose to share the postgresql UNIX socket
```yml
# Runs SQL LRS with Postgres - Provided for demonstration purposes only!
# To run: docker compose up
# See the Docker Compose docs for more info: https://docs.docker.com/compose/
services:
  db:
    image: postgres
    volumes:
      - ./tmp/db:/var/lib/postgresql/data
      - ./postgres/socket:/var/run/postgresql
    environment:
      POSTGRES_USER: lrsql_user
      POSTGRES_PASSWORD: lrsql_password
      POSTGRES_DB: lrsql_db
  lrs:
    # build: . # switch to this for active dev
    image: yetanalytics/lrsql:latest
    command:
      - /lrsql/bin/run_postgres.sh
    ports:
      - “8080:8080”
    depends_on:
      - db
    volumes:
      - ./postgres/socket:/var/run/postgresql
    environment:
      LRSQL_API_KEY_DEFAULT: my_key
      LRSQL_API_SECRET_DEFAULT: my_secret
      LRSQL_ADMIN_USER_DEFAULT: my_username
      LRSQL_ADMIN_PASS_DEFAULT: my_password
[...]

PR #239, opened 2022-07-08
SQL-161 add unix socket support

milt:
[SQL-161] Per https://github.com/yetanalytics/lrsql/issues/235 we'd like to be able to support the use of unix sockets with [junixsocket](https://kohlschutter.github.io/junixsocket/dependency.html).

This PR adds junixsocket at version 2.6.1 and provides some documentation on usage.

[SQL-161]: https://yet.atlassian.net/browse/SQL-161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kelvinqian00:
Is it possible to add a test for this?

milt:
> Is it possible to add a test for this?

It would be possible to test using testcontainers + docker compose I suppose, but it's a lot of work to just test the underlying lib which is already tested.

kelvinqian00:
> It would be possible to test using testcontainers + docker compose I suppose, but it's a lot of work to just test the underlying lib which is already tested.

Okay, even then it should be simple enough to test manually by passing in that JDBC URL, which I'm assuming you already did on your end.

milt:
Yep, in the course of troubleshooting https://github.com/yetanalytics/lrsql/issues/235

kelvinqian00 (approved):

PR #246, opened 2022-08-18
OIDC and Authority Template security and validation

kelvinqian00:
- Add validation to default non-OIDC authority template and all OIDC authority templates
  - OIDC templates must conform to 3-legged OAuth xAPI Groups
- Fix bug/security hole in `authorize-oidc-request` interceptor where `authorize-admin-action`'s result is always interpreted as `true`.


kelvinqian00:
Will perform the two main tasks in this PR separately (see #247).

PR #247, opened 2022-08-18
OIDC security fix

kelvinqian00:
Fix bug/security hole in `authorize-oidc-request` interceptor where `authorize-admin-action`'s result is always interpreted as true.


milt (approved):
Good find!

PR #249, opened 2022-08-25
statements/read/mine OIDC demo

milt:
* Updates all OIDC keycloak demo uris to use `localhost` instead of `0.0.0.0`
  * Makes clients that don't tolerate that foolishness work
* Adds `lrs:statements/read/mine` scope to keycloak demo client scopes
* Replaces `lrs:all` default scope in `lrs_client` Client Credentials Grant demo with `lrs:statements/read/mine` and `lrs:statements/write`
 
To test:
1. `make keycloak-demo`
2. `make ephemeral-oidc`
3. In an Oauth 2.0 CCG-capable http client, use the client: `lrs_client`. Get its secret from the [keycloak UI](http://localhost:8081/auth/admin/master/console/). For the token URL, use `http://localhost:8081/auth/realms/test/protocol/openid-connect/token`

### DIFF
--- a/dev-resources/keycloak_demo/test-realm.json
+++ b/dev-resources/keycloak_demo/test-realm.json
@@ -545,14 +545,14 @@
     "id" : "93f48f9d-bd78-44f2-97f7-7f3b65aaa8da",
     "clientId" : "lrs",
     "name" : "SQL LRS",
-    "rootUrl" : "http://0.0.0.0:8080/",
-    "adminUrl" : "http://0.0.0.0:8080/",
+    "rootUrl" : "http://localhost:8080/",
+    "adminUrl" : "http://localhost:8080/",
     "surrogateAuthRequired" : false,
     "enabled" : true,
     "alwaysDisplayInConsole" : false,
     "clientAuthenticatorType" : "client-secret",
-    "redirectUris" : [ "http://0.0.0.0:8080/*" ],
-    "webOrigins" : [ "http://0.0.0.0:8080" ],
+    "redirectUris" : [ "http://localhost:8080/*" ],
+    "webOrigins" : [ "http://localhost:8080" ],
     "notBefore" : 0,
     "bearerOnly" : true,
     "consentRequired" : false,
@@ -602,8 +602,8 @@
     "enabled" : true,
     "alwaysDisplayInConsole" : false,
     "clientAuthenticatorType" : "client-secret",
-    "redirectUris" : [ "http://localhost:9500/", "http://0.0.0.0:8080/admin/index.html", "http://localhost:9500/*" ],
-    "webOrigins" : [ "http://localhost:9500", "http://0.0.0.0:8080" ],
+    "redirectUris" : [ "http://localhost:9500/", "http://localhost:8080/admin/index.html", "http://localhost:9500/*" ],
+    "webOrigins" : [ "http://localhost:8080", "http://localhost:9500" ],
     "notBefore" : 0,
     "bearerOnly" : false,
     "consentRequired" : false,
@@ -653,7 +653,7 @@
       "config" : {
         "id.token.claim" : "false",
         "access.token.claim" : "true",
-        "included.custom.audience" : "http://0.0.0.0:8080",
+        "included.custom.audience" : "http://localhost:8080",
         "userinfo.token.claim" : "false"
       }
     } ],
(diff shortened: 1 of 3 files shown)


dev-resources/keycloak_demo/test-realm.json
```diff
@@ -1372,7 +1381,7 @@
       "subType" : "anonymous",
       "subComponents" : { },
       "config" : {
-        "allowed-protocol-mapper-types" : [ "saml-user-property-mapper", "oidc-address-mapper", "saml-role-list-mapper", "oidc-usermodel-attribute-mapper", "oidc-sha256-pairwise-sub-mapper", "oidc-full-name-mapper", "oidc-usermodel-property-mapper", "saml-user-attribute-mapper" ]
+        "allowed-protocol-mapper-types" : [ "saml-user-attribute-mapper", "oidc-full-name-mapper", "oidc-usermodel-attribute-mapper", "oidc-usermodel-property-mapper", "oidc-address-mapper", "oidc-sha256-pairwise-sub-mapper", "saml-user-property-mapper", "saml-role-list-mapper" ]
```
kelvinqian00, line 1384: What's going on here? It seems that all you did here was reorder the list entries.

dev-resources/keycloak_demo/test-realm.json
```diff
@@ -1252,6 +1252,15 @@
...
     }
+  }, {
+    "id" : "0448c583-b1f9-4cc5-90fd-b7c591966e69",
+    "name" : "lrs:statements/read/mine",
+    "description" : "Read own statements",
+    "protocol" : "openid-connect",
+    "attributes" : {
+      "include.in.token.scope" : "true",
+      "display.on.consent.screen" : "true"
+    }
```
kelvinqian00, line 1263: I see that this scope was added in addition to `lrs:all/read`. Wouldn't `lrs:all/read` overwrite the new scope though?

dev-resources/keycloak_demo/test-realm.json
```diff
@@ -763,7 +763,7 @@
         "jsonType.label" : "String"
       }
     } ],
-    "defaultClientScopes" : [ "web-origins", "profile", "roles", "lrs:all", "email" ],
+    "defaultClientScopes" : [ "web-origins", "profile", "roles", "lrs:statements/write", "lrs:statements/read/mine", "email" ],
```
kelvinqian00, line 766: I wonder what the implications would be for narrowing the default client scopes. (If it just affects the demo then I expect not many.)

dev-resources/keycloak_demo/test-realm.json
milt: that probably happened on the re-export from keycloak

dev-resources/keycloak_demo/test-realm.json
milt: this is the client scope itself. What applies is the `defaultClientScopes` of the client, `lrs`

dev-resources/keycloak_demo/test-realm.json
kelvinqian00: Yeah that would make sense if the new realm file was auto-genned (imagine if you had to manually edit all the components!)

dev-resources/keycloak_demo/test-realm.json
milt: Yeah, it's just the demo, and this demonstrates OIDC better (I can only read/write my own statements)

dev-resources/keycloak_demo/test-realm.json
milt: Actually for the 0.0.0.0 -> localhost I did just that, but then there were more significant changes so I just used the UI

dev-resources/keycloak_demo/test-realm.json
kelvinqian00: Wouldn't you also need `lrs:statements/write` since you added that to the default scopes?

milt: it's in there https://github.com/yetanalytics/lrsql/blob/37c779e7ac0cf00f81a3c40794db58764a995574/dev-resources/keycloak_demo/test-realm.json#L1157-L1167

kelvinqian00 (approved):

PR #251, opened 2022-10-24
NVD Clojure version update

kelvinqian00:


kelvinqian00:
Closed since updating the nvd-clojure version did not uncover a CVE in this instance.

PR #259, opened 2022-12-08
AWS template fixes

cliffcaseyyet:


dev-resources/template/2_lrs.yml
```diff
@@ -408,6 +408,8 @@ Resources:
                 Description=SQL LRS Service
                 [Service]
                 User=root
+                # Required to prevent arbitrary encoding settings
+                Environment="LC_ALL=en_US.UTF-8"
```
milt, line 412: rereading this I wonder if we shouldn't also set LANG to the same thing

milt (approved):

PR #260, opened 2022-12-08
[SQL-157] Encoding fix

kelvinqian00:
Add/modify util functions that encode or decode strings such that the charset is always fixed to UTF-8, and replace all instances of `String` constructor calls to use said util functions. Also add tests on Unicode characters.

~TODO: However, these fixes are not enough to address the issues causing the SQL-157 Out of Memory Error bug caused by incorrect encoding. This will have to wait until the `lrs` library can be updated.~

Turns out updating the `lrs` lib is impractical due to the default charset being used deep within Pedestal (see [here](https://github.com/pedestal/pedestal/issues/582)). Thus we instead set the default encoding using Java variables in our shell scripts and executable configs.


milt:
don't forget https://github.com/yetanalytics/lrsql/blob/encoding-fix/src/main/lrsql/util/statement.clj#L129

kelvinqian00:
> don't forget https://github.com/yetanalytics/lrsql/blob/encoding-fix/src/main/lrsql/util/statement.clj#L129

191fb5c0d009a38ffe4d6718ece997da00802b40

PR #262, opened 2022-12-13
Miscellaneous updates

kelvinqian00:
Some things I wanted to fix before the upcoming break ver is released
- Fix docstrings and formatting
- Remove some outdated TODOs
- Replace `(dorun (map ...))` with `run!`
- Use more util functions in place of Java interop
- Patch reflection warnings
- Enable (conn pool) JMX monitoring by default for test (not prod) mode


resources/lrsql/config/test/default/connection.edn
```diff
@@ -11,4 +11,4 @@
  :pool-leak-detection-threshold    0
  :pool-transaction-isolation       nil
  :pool-name               nil
- :pool-enable-jmx         false}
+ :pool-enable-jmx         true}
```
milt, line 14: Isn't this just JMX for the conn pool? the description made it seem like it was across the board. Also if the benchmarks use the test profile I'm pretty sure you'll want all JMX off

kelvinqian00: Yes I should've specified that in the description. In fact I was motivated to turn this on since JMX was turned on for other classes, such as Pedestal.

I didn't realize it might affect benching though; if that's true it would be a good reason to revert the change (and add a comment in the config file). I do wonder why it's JMX on the conn pool, and not on other classes, that'll affect this.

src/main/lrsql/ops/command/auth.clj
```diff
@@ -32,7 +32,7 @@
   "Insert `input`, a seq of maps where each API key pair is associated
    with a different scope."
   [bk tx input]
-  (dorun (map (partial bp/-insert-credential-scope! bk tx) input)))
+  (run! (partial bp/-insert-credential-scope! bk tx) input))
```
milt, line 35: TIL

milt (approved):

resources/lrsql/config/test/default/connection.edn
kelvinqian00: > Also if the benchmarks use the test profile I'm pretty sure you'll want **all** JMX off

Oh didn't see that. Given that all benching has been done with the "some JMX is on" configuration I've talked about, I'm not sure if adding some to the conn pool will affect things that much.

PR #265, opened 2023-01-03
CORS Configuration Support

cliffcaseyyet:
This enables CORS Support to be configurable. There are three possible running modes, defined by 2 variables:

- By default, with no additional settings, the allowed origins will be restricted to the `LRSQL_HTTP_HOST` setting, which is by default `0.0.0.0:[8080/8443]`
- If you set `LRSQL_ALLOW_ALL_ORIGINS` to `true`, it will ignore and just approve everything (legacy mode)
- If you explicitly set origins using `LRSQL_ALLOWED_ORIGINS` it will restrict to those. in ENV thats a comma-separated list, in JSON its an array.

### DIFF
--- a/resources/lrsql/config/test/default/webserver.edn
+++ b/resources/lrsql/config/test/default/webserver.edn
@@ -9,6 +9,8 @@
  :ssl-port                  8443
  :http-host                 "0.0.0.0"
  :http-port                 8080
+ :allow-all-origins         false
+ :allowed-origins           nil
  :url-prefix                "/xapi"
  :enable-admin-ui           true
  :enable-stmt-html          true
--- a/src/main/lrsql/system.clj
+++ b/src/main/lrsql/system.clj
@@ -9,9 +9,13 @@
 (defn system
   "Return a lrsql system with configuration specified by the `profile`
    keyword."
-  [backend profile]
-  (let [config
+  [backend profile & {:keys [conf-overrides]}]
+  (let [profile-config
         (read-config profile)
+        config (reduce (fn [agg [key val]]
+                         (assoc-in agg key val))
+                       profile-config
+                       (seq conf-overrides))
         initial-sys ; init without configuration
         (component/system-map
          ;; Logger is required by backend so it happens first
--- a/src/main/lrsql/system.clj
+++ b/src/main/lrsql/system.clj
@@ -8,7 +8,8 @@
 
 (defn system
   "Return a lrsql system with configuration specified by the `profile`
-   keyword."
+   keyword. Optional kwarg ``conf-overrides`` takes a map where the key is a vec
+   of a key location in config map, and the value is an override."
   [backend profile & {:keys [conf-overrides]}]
   (let [profile-config
         (read-config profile)
--- a/src/main/lrsql/system/webserver.clj
+++ b/src/main/lrsql/system/webserver.clj
@@ -76,11 +76,10 @@
      {:creds           true
       :allowed-origins (fn [origin]
                          (or allow-all-origins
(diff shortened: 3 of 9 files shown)


src/main/lrsql/system/webserver.clj
```diff
@@ -71,7 +74,12 @@
      ::i/enable-statement-html enable-stmt-html
      ::http/allowed-origins
      {:creds           true
-      :allowed-origins (constantly true)}
+      :allowed-origins (fn [origin]
+                         (or allow-all-origins
+                             (some #(= origin %)
+                                   (or allowed-origins
+                                       [(format "http://%s:%s" http-host http-port)
+                                        (format "https://%s:%s" http-host ssl-port)]))))}
```
kelvinqian00, line 82: This should be extracted into a helper function (and also maybe extract the array of formatted IRIs into its own `let` binding)

resources/lrsql/config/test/default/webserver.edn
```diff
@@ -9,6 +9,8 @@
  :ssl-port                  8443
  :http-host                 "0.0.0.0"
  :http-port                 8080
+ :allow-all-origins         false
+ :allowed-origins           nil
```
kelvinqian00, line 13: I wonder if you can actually set `allowed-origins` to some array and then test the CORS setting using a `babshka.curl` request.

src/main/lrsql/system/webserver.clj
kelvinqian00: Or in general just make these lines of code a bit less horrifying.

resources/lrsql/config/test/default/webserver.edn
cliffcaseyyet: see https://github.com/yetanalytics/lrsql/pull/265/commits/06d2bfc5116eb0f8b6e928a993bcd4f7bb02939e

src/main/lrsql/system.clj
```diff
@@ -9,9 +9,13 @@
 (defn system
   "Return a lrsql system with configuration specified by the `profile`
    keyword."
```
kelvinqian00, line 11: Make sure to document the new `conf-overrides` arg in the docstring.

src/test/lrsql/admin/cors_test.clj
```diff
@@ -0,0 +1,120 @@
...
+    (testing "create account with default CORS check success"
+      (let [good-cors-headers
+            (merge headers {"Origin" "http://0.0.0.0:8080"})
+            {:keys [status body] :as response}
+            (create-account good-cors-headers req-body)
+            edn-body       (u/parse-json body)]
+        (is (= 200 status))
+        (is (try (u/str->uuid (get edn-body "account-id"))
+                 (catch Exception _ false)))))
+    (component/stop sys')))
```
kelvinqian00, line 82: I'd be tempted to refactor out this `testing` code into a custom macro - not only would it deduplicate code, but it would also emphasize how similar the two tests are, in which everything is the same except that you switch which origins are allowed or not.

src/main/lrsql/system.clj
cliffcaseyyet: done

src/test/lrsql/admin/cors_test.clj
cliffcaseyyet: discussed offline

src/main/lrsql/system/webserver.clj
cliffcaseyyet: done

kelvinqian00 (approved):

PR #268, opened 2023-01-13
SQL-163 localhost-friendly CORS defaults

milt:
[SQL-163] With no CORS allowed origins specified, `localhost` should be a valid origin. This PR adds it to the default list with the given HTTP/S ports.

[SQL-163]: https://yet.atlassian.net/browse/SQL-163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/lrsql/system/webserver.clj
+++ b/src/main/lrsql/system/webserver.clj
@@ -65,7 +65,9 @@
         ;; default ports
         allowed-list
         (or allowed-origins
-            (cond-> [(format "http://%s:%s" http-host http-port)
+            (cond-> [(format "http://localhost:%s" http-port)
+                     (format "https://localhost:%s" ssl-port)
+                     (format "http://%s:%s" http-host http-port)
                      (format "https://%s:%s" http-host ssl-port)]
               (= http-port 80) (conj (format "http://%s" http-host))
               (= ssl-port 443) (conj (format "https://%s" http-host))))]
--- a/src/test/lrsql/admin/cors_test.clj
+++ b/src/test/lrsql/admin/cors_test.clj
@@ -23,13 +23,13 @@
 
 (defn- login-account
   [headers body]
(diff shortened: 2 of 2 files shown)


src/test/lrsql/admin/cors_test.clj
```diff
@@ -72,7 +72,7 @@
         (is-err-code (create-account bad-cors-headers req-body) 403)))
     (testing "create account with default CORS check success"
       (let [good-cors-headers
-            (merge headers {"Origin" "http://0.0.0.0:8080"})
+            (merge headers {"Origin" "http://localhost:8080"})
```
kelvinqian00, line 75: I'm assuming that you're overriding the existing tests, as opposing to adding new tests, since all places where `0.0.0.0` works should also work with `localhost`, right?

src/main/lrsql/system/webserver.clj
```diff
@@ -65,7 +65,9 @@
         ;; default ports
         allowed-list
         (or allowed-origins
-            (cond-> [(format "http://%s:%s" http-host http-port)
+            (cond-> [(format "http://localhost:%s" http-port)
+                     (format "https://localhost:%s" ssl-port)
```
kelvinqian00, line 69: And by the same principle, there's no need to add `http:0.0.0.0:%s" as one of the default allowed origins.

milt: Since that is the default bind host, not needed yeah

src/test/lrsql/admin/cors_test.clj
milt: yep! Figured localhost is just better from an instructive standpoint

kelvinqian00 (approved):

cliffcaseyyet (approved):

PR #282, opened 2023-02-07
SQL-172 Admin LRS Monitor Dashboard

milt:
[SQL-172] Add a screen to the admin UI with simple vis of system-wide xAPI activity.

Feed it with a simple API that returns JSON. The API can be toggled on/off with a configuration variable.

API currently has a single endpoint `/admin/status` that returns a JSON object with the following properties:

- [x] `statement-count` - count of statements in the LRS
- [x] `actor-count` - count of distinct statement actors
- [x] `last-statement-stored` - last statement stored timestamp
- [x] `platform-frequency` - map of platform string (or unknown) to count of statements with that `context.platform`
- [x] `timeline` -  configurable timeline of statements submitted by `stored`

Desired vis is specified through the `include` param that can be passed multiple times for batch vis.

[SQL-172]: https://yet.atlassian.net/browse/SQL-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


doc/env_vars.md
```diff
@@ -119,6 +119,7 @@ You may have noted that some options are not available:
 | `LRSQL_ALLOWED_ORIGINS` | `allowedOrigins` | This is a list of allowed origins which overrides the defaults. As an ENV it should be written as a comma separated list with no spaces. In JSON config it can be written directly as an array, e.g. `["http://myhost:myport", "https://anotherhost"]` | Not Set |
 | `LRSQL_URL_PREFIX` | `urlPrefix` | The prefix of the webserver URL path, e.g. the prefix in `http://0.0.0.0:8080/xapi` is `/xapi`. Used when constructing the `more` value for multi-statement queries. | `/xapi` |
 | `LRSQL_ENABLE_ADMIN_UI` | `enableAdminUi` | Whether or not to serve the administrative UI at `/admin` | `true` |
+| `LRSQL_ENABLE_ADMIN_STATUS` | `enableAdminStatus` | Whether or not to serve admin system status data that queries the database. | `true` |
```
kelvinqian00, line 122: We really need to break up the "Webserver" config var table at this point.

src/main/lrsql/system/lrs.clj
```diff
@@ -328,4 +328,10 @@
       (jdbc/with-transaction [tx conn]
         (auth-cmd/delete-credential! backend tx cred-in)
         {:api-key    api-key
-         :secret-key secret-key}))))
+         :secret-key secret-key})))
+  adp/AdminStatusProvider
```
kelvinqian00, line 332: Should have an extra space

kelvinqian00 (approved):

src/main/lrsql/system/lrs.clj
milt: https://github.com/yetanalytics/lrsql/pull/282/commits/44b6d23089d8bcde98d2cc2832024436bf8139ba

src/db/postgres/lrsql/postgres/sql/query.sql
```diff
@@ -287,3 +287,50 @@ AND secret_key = :secret-key;
...
+SELECT payload->>'stored' lstored
+FROM xapi_statement
+ORDER BY id DESC
+LIMIT 1;
+
+-- :name query-platform-frequency
+-- :command :query
+-- :result :many
+-- :doc Return counts of platforms used in statements.
+SELECT COALESCE(payload#>>'{context,platform}', 'unknown') platform,
```
kelvinqian00, line 321: Huh apparently GitHub does not like this

src/main/lrsql/admin/interceptors/ui.clj
```diff
@@ -9,20 +9,26 @@
   [_]
   (resp/redirect "/admin/index.html"))
 
-(def get-env
-  "Provide select config data to client upon request"
+(defn get-env
+  "Provide select config data to client upon request. Takes a map with static
+  config to inject:
+    :enable-admin-status - boolean, determines if the admin status endpoint is
+      enabled."
```
kelvinqian00, line 16: Weird indentation?

src/main/lrsql/admin/interceptors/ui.clj
```diff
@@ -9,20 +9,26 @@
...
+  [{:keys [enable-admin-status]
+    :or   {enable-admin-status false}}]
```
kelvinqian00, line 18: More broadly, is there a reason why you're passing in `enable-admin-status` as a function arg, instead of pulling it directly from `ctx` like the existing config values?

src/main/lrsql/spec/admin/status.clj
```diff
@@ -0,0 +1,114 @@
...
+      nat-int?))
+
+(s/def :lrsql.spec.admin.status.params/timeline-since
+  (s/and ::xs/timestamp
+         after-epoch?))
+(s/def :lrsql.spec.admin.status.params/timeline-until
+  (s/and ::xs/timestamp
+         after-epoch?))
+
+
```
kelvinqian00, line 59: Extra space? (Also there's some wonky spacing in the rest of this namespace)

src/main/lrsql/system/lrs.clj
```diff
@@ -328,4 +329,12 @@
...
-         :secret-key secret-key}))))
+         :secret-key secret-key})))
+
+  adp/AdminStatusProvider
+  (-get-status
+    [this params]
+    (let [conn (lrs-conn this)]
+      (jdbc/with-transaction [tx conn]
+        (admin-q/query-status
+         backend tx (admin-stat-input/query-status-input params))))))
```
kelvinqian00, line 340: Would probably be clearer if the input fn call was put in a let-binding.

src/test/lrsql/admin/route_test.clj
```diff
@@ -190,6 +195,26 @@
...
+                "platform-frequency"    {}
+                "timeline"              []}
+               edn-body)))
+      (testing "validates params"
+        (let [{:keys [status]} (curl/get
+                                "http://localhost:8080/admin/status?timeline-unit=foo"
+                                {:headers headers
+                                 :throw   false})]
+          ;; failure
+          (is (= 400 status)))))
```
kelvinqian00, line 217: It would be helpful to have a test case with a valid param, just so we can compare it with the invalid case.

src/main/lrsql/ops/query/admin.clj
```diff
@@ -61,3 +63,42 @@
...
+            (fn [m {:keys [platform scount]}]
+              (assoc m platform scount))
+            {}
+            (bp/-query-platform-frequency bk tx)))
+
+    (include "timeline")
+    (assoc :timeline
+           (mapv
+            (fn [{:keys [stored scount]}]
+              {:stored (u/pad-time-str stored)
```
kelvinqian00, line 102: Is there a reason why we need to pad the timestamp string here?

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -130,3 +149,33 @@
...
+             (adp/-get-status lrs {})))
+      ;; add a statement
+      (lrsp/-store-statements lrs auth-ident [stmt-0] [])
+      (let [last-stored (get-in
+                         (lrsp/-get-statements lrs auth-ident {} [])
+                         [:statement-result
+                          :statements
+                          0
+                          "stored"])]
+        (is (= {:statement-count       1
```
kelvinqian00, line 172: I also feel like it would be helpful to have a test on more than one statement. For example, the second statement could share the same actor or platform as the previous, demonstrating that the values are not coupled with each other.

kelvinqian00:
Also, as always, don't forget documentation.

src/db/h2/lrsql/h2/result.clj
```diff
@@ -0,0 +1,41 @@
...
+    {:lstored (-> id squuid/uuid->time u/time->str)}))
+
+(defn query-platform-frequency-result
+  "Given a result of all statement payloads, return the frequencies of each"
+  [query-result]
+  (map
+   (fn [[platform scount]]
+     {:platform platform
+      :scount   scount})
+   (frequencies
```
kelvinqian00, line 20: This is a core function? 🤯

src/db/h2/lrsql/h2/result.clj
```diff
@@ -0,0 +1,41 @@
...
+    (map #(get-in % [:payload "context" "platform"] "unknown")
```
kelvinqian00, line 21: Is "unknown" really the right word to use in the case the statement has no platform (as opposed to "none" or similar)?

src/main/lrsql/input/admin/status.clj
```diff
@@ -0,0 +1,51 @@
...
+  [{:keys [timeline-unit
+           timeline-since
+           timeline-until]
+    :or   {timeline-unit "day"}}]
+  (let [since' (or (some-> timeline-since u/str->time)
+                   (u/offset-time (u/current-time) -14 :days))
+        until' (or (some-> timeline-until u/str->time)
+                   (u/current-time))]
+    {:unit-for (get unit-for timeline-unit)
+     :since-id (squuid/time->uuid since')
```
kelvinqian00, line 31: Just be careful about how the underlying `squuid/time->uuid` works:
```clojure
(time->uuid #inst "2021-12-22T14:33:04.769000000-00:00")
;; => #uuid "017de28f-5801-8fff-8fff-ffffffffffff"
```
Notice how the random digits are all maxed out. Meaning that if you have a randomly generated SQUUID with the same time bits but with random bits less than the maximum, it will fall outside of the range. So you'd want to decrement the `since'` timestamp by a nanosecond/millisecond (whichever is the smallest time interval).

src/main/lrsql/spec/admin/status.clj
```diff
@@ -0,0 +1,114 @@
...
+(def get-status-params-spec
+  (s/keys :opt-un [:lrsql.spec.admin.status.params/include
+                   :lrsql.spec.admin.status.params/timeline-unit
+                   :lrsql.spec.admin.status.params/timeline-since
+                   :lrsql.spec.admin.status.params/timeline-until]))
```
kelvinqian00, line 64: You might want to add restrictions on if the `timeline-unit` covers an interval greater than the interval between `timeline-since` and `timeline-util` (e.g. you choose year as your unit but your since-to-until interval only covers two months). Right now in the UI nothing shows up in the timeline in that case and it feels pretty weird.

src/db/postgres/lrsql/postgres/sql/query.sql
milt: Yeah it is not uncommon for GH syntax highlighting to break. To be fair SQL syntax is almost impossible to predict across implementations and it won't get better until they delegate it to an LLM or something.

src/main/lrsql/admin/interceptors/ui.clj
milt: I made that choice because it seemed better to do all the work at init (since it turns off a route) rather than adding another interceptor just to stick a value in at runtime. If there was a single interceptor for sticking static config into the CTX at runtime I might have made a different call, but I couldn't find one so I chose this.

src/main/lrsql/admin/interceptors/ui.clj
milt: I indented 2 spaces so it is clear it is under the list item `:enable-admin-status`. If there are multiple items in the future is clearer IMO

src/main/lrsql/spec/admin/status.clj
milt: Do you mean extra line break? Sure I can remove it.

milt: https://github.com/yetanalytics/lrsql/pull/282/commits/be2d844fa26c37d66531a41901d1e8494cd03605

src/main/lrsql/system/lrs.clj
milt: Yeah sure, reading the other methods that would be consistent. https://github.com/yetanalytics/lrsql/pull/282/commits/5a60033925e951a2e317a74f417b576be00ce905

src/test/lrsql/admin/route_test.clj
milt: _Technically_ it has one, since the lack of params is a valid params map `{}` but sure.

milt: https://github.com/yetanalytics/lrsql/pull/282/commits/c09ecbe58825f4f65b0b64c11631b27d50163fa3

src/main/lrsql/ops/query/admin.clj
milt: Yes. Since timestamps are ISO zulu (and normalized) in the DB, the timeline query uses a substring call to bucket them. This results in a partial timestamp which needs to have the rest (zeroed) re-added.

src/test/lrsql/admin/protocol_test.clj
milt: I agree, BUT: Given the nature of the timeline query doing this with 2 statements technically introduces a false test fail that could happen (at the very least) once per year. 😸 
I'll see what I can do.

src/main/lrsql/admin/interceptors/ui.clj
kelvinqian00: Hmm, I actually undid the 2-spaces indentation in other comments because I thought it looked ugly. I had no idea there was an actual reason for that, sorry.

src/test/lrsql/admin/protocol_test.clj
kelvinqian00: Why would that happen? Daylights Savings Time?

milt: If the unit is the default (day), you could technically have a case where the two stamps could have differing days.

milt: What I do here should prevent it from ever happening: https://github.com/yetanalytics/lrsql/pull/282/commits/9f3dd96843e62bb1756dfdfded6a242e0c7e79ac

src/db/h2/lrsql/h2/result.clj
milt: Haha yeah the usual philosophy of keeping the core small can take a backseat to things that are just hugely useful to the working programmer and I'm here for it.

src/main/lrsql/admin/interceptors/ui.clj
kelvinqian00: I wonder how all the other env vars in that function/interceptor _do_ get injected into the context then. Feels weird that we'd introduce some on init and others at runtime.

src/db/h2/lrsql/h2/result.clj
milt: Good question. I came at it from the position that every statement _comes from a platform of some kind_ but maybe that's wrong. I'll read the spec language about it.

src/main/lrsql/ops/query/admin.clj
kelvinqian00: Hmm so all the timestamp values used in the lrsql app _are_ already normalized, and we'd just have to convert queried timestamps to it?

src/db/h2/lrsql/h2/result.clj
kelvinqian00: If only I knew, would've been useful for, say, ID counting in Pan.

src/main/lrsql/admin/interceptors/ui.clj
milt: No worries, there really isn't an authoritative source on it so 🤷

src/main/lrsql/admin/interceptors/ui.clj
milt: Two of them come from the LRS lib (path-prefix and statement-html?) and are added for runtime because they are used in a variety of places. The OIDC env is added by an interceptor because it is also required at runtime I believe, but if it is just being injected for this it could be moved into a static arg (though that would be outside the scope of this PR)

src/main/lrsql/ops/query/admin.clj
milt: Not sure exactly what you mean, but yeah I think so? Basically what you'd get back for instance for the `hour` unit would be `2023-02-22T12` and this would make it `2023-02-22T12:00:00.000000000Z` so it is valid and normalized.

src/main/lrsql/input/admin/status.clj
milt: > whichever is the smallest time interval

Do you mean whichever is the smallest time interval the Instant API allows?

src/main/lrsql/ops/query/admin.clj
kelvinqian00: I basically mean timestamp strings that are already used in the app layer, though it seems that they're generally used as Instants and not strings. That said there are other places where timestamp strings are returned, but _they_ don't get padded out, so I'd imagine this padding is required by the UI.

src/main/lrsql/input/admin/status.clj
kelvinqian00: Yeah...now that I recall though, the colossal-squuid library uses milliseconds as the smallest time interval for its purposes, so go with that.

src/main/lrsql/spec/admin/status.clj
milt: I'll look into this on the front-end. On the back end I think it's just GIGO that it doesn't make sense to spend the time to police.

src/main/lrsql/ops/query/admin.clj
kelvinqian00: Places where timestamp strings are returned, but without calling `pad-time-str` first: https://github.com/yetanalytics/lrsql/blob/d53cb5ce53d352730694a57db8abf8cdcdc3ae9b/src/main/lrsql/ops/query/document.clj#L42

src/db/h2/lrsql/h2/result.clj
milt: [According to the spec](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#requirements-10) platform only gets supplied for statements with an activity object, so not every statement, though I don't think we need to really differentiate in the query result.
I think `none` is fine I'll go with that.

kelvinqian00: Cool. As a side note there could be issues if there's a platform literally named `none` (or previously `unknown`) but nothing to do about that. :shrug:

src/main/lrsql/input/admin/status.clj
milt: https://github.com/yetanalytics/lrsql/pull/282/commits/64d17f39ba3c5f3189eb023c7dd2d6b2c04191b6

milt:
Documentation is pending frontend approval

src/db/h2/lrsql/h2/result.clj
milt: True, though the meaning is the same either way, explicit or implicit haha

src/main/lrsql/ops/query/admin.clj
kelvinqian00: As discussed during standup, I'd add a comment where `pad-time-str` is used to clarify why it's used here and not in other places where timestamp strings are returned.

milt: https://github.com/yetanalytics/lrsql/pull/282/commits/f410659efe8b3c568e275e0ef419ff29be0e8849

cliffcaseyyet (approved):

kelvinqian00 (approved):

PR #288, opened 2023-02-27
SQL-177 Test against all supported versions of PostgreSQL

milt:
[SQL-177] We support all current maintained versions of PG but do not test them in CI. This PR adds targets to the makefile to explicitly test each version. Note that the existing `test-postgres` target will run the earliest version, 11.

Note that the _current minor_ version of each major version will run.

The `LRSQL_TEST_DB_VERSION` env var determines which version will be run in TestContainers. Any Postgres docker tag (ex. `latest`) is valid. The default is `11`.

[SQL-177]: https://yet.atlassian.net/browse/SQL-177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/resources/lrsql/config/config.edn
+++ b/resources/lrsql/config/config.edn
@@ -22,7 +22,7 @@
                    ;; :db-properties "currentSchema=lrsql"
 
                    ;; Testing Only! Specify the version used with testcontainers
-                   :test-db-version "11.9"}
+                   :test-db-version #or [#env LRSQL_TEST_DB_VERSION "11"]}
    :test-oidc     {:db-type "h2:mem"
                    :db-name "ephemeral"}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,10 +17,17 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
-  
+
     strategy:
       matrix:
-        target: [test-h2, test-sqlite, test-postgres]
+        target:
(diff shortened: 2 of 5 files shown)


resources/lrsql/config/config.edn
```diff
@@ -22,7 +22,7 @@
                    ;; :db-properties "currentSchema=lrsql"
 
                    ;; Testing Only! Specify the version used with testcontainers
-                   :test-db-version "11.9"}
+                   :test-db-version #or [#env LRSQL_TEST_DB_VERSION "11"]}
```
kelvinqian00, line 25: Do we need to document this env var? We do have _some_ documentation touching upon dev-only stuff (e.g. benching), so I'd say it's a valid question.

milt: I can put something in the `dev.md` file about it sure. Don't want to put it in the main env var doc because it isn't runtime.

milt: https://github.com/yetanalytics/lrsql/pull/288/commits/4d62b4d0ae9baa44323d810bd3133844db9c2409

kelvinqian00 (approved):

cliffcaseyyet (approved):
love it. long time coming.

PR #289, opened 2023-03-01
SQL-179 Use rds-db-init for db init function in CF template

milt:
[SQL-179] Rather than deploying the RDS db user init function from this project, this PR switches the template to use [rds-db-init](https://github.com/yetanalytics/rds-db-init).

- [x] Remove db-init-script
- [x] Update CF template
- [x] Document override procedures for regions outside the US

[SQL-179]: https://yet.atlassian.net/browse/SQL-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dev-resources/template/2_lrs.yml
```diff
@@ -142,11 +147,14 @@ Resources:
     Type: AWS::Lambda::Function
     Properties:
       Code:
-        S3Bucket: !Ref S3Bucket
+        S3Bucket: !If
+          - SetDBInitFnBucket
+          - !Ref DBInitFnBucketOverride
+          - !Sub "yet-rds-db-init-deploy-${AWS::Region}"
```
kelvinqian00, line 153: Do you know what would happen if you upload to a non-US region _without_ setting `DBInitFnBucketOverride` or `SetDBInitFnBucket`? Will it fail gracefully or will the whole thing explode?

dev-resources/template/2_lrs.yml
```diff
@@ -142,11 +147,14 @@ Resources:
...
         S3Key: !If
-          - SetS3Key
-          - !Ref S3KeyOverride
-          - !Sub "db-init-${LrsVersion}.zip"
+          - SetDBInitFnKey
+          - !Ref DBInitFnKeyOverride
+          - !Sub "rds-db-init-v${DBInitFnVersion}.zip"
```
kelvinqian00, line 157: Likewise the same question but for `DBInitFnKeyOverride` and `SetDBInitFnKey`.

dev-resources/template/2_lrs.yml
milt: So you're asking what happens if you upload the DB function to a bucket in your region but don't set this? It will do what the template does currently, which is give a descriptive error saying you can't use a bucket in another region.

kelvinqian00: Ok so it fails gracefully.

milt: Actually it is a bit different. It will say that it can't find a bucket _in your region_ with the name `yet-rds-db-init-deploy-<region name>`, but yeah as graceful as any other CloudFormation failure. Specification of more specific errors would be a nice CF feature but not one that exists.

dev-resources/template/2_lrs.yml
```diff
@@ -142,11 +147,14 @@ Resources:
     Type: AWS::Lambda::Function
     Properties:
       Code:
-        S3Bucket: !Ref S3Bucket
+        S3Bucket: !If
```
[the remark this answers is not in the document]
milt: huh? Do you mean at runtime? They're not passed to the function itself at runtime. They are given as the _code location_ for the lambda function itself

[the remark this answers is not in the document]
milt: Also those specific params don't exist anymore (in this PR)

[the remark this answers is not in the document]
milt: The `Code` property of `AWS::Lambda::Function`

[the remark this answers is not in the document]
kelvinqian00: Oh so by "code location" you mean the actual S3 bucket/file that contains the lambda function.

kelvinqian00 (approved):

.github/workflows/build.yml
```diff
@@ -142,31 +142,3 @@ jobs:
...
-
-      - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1-node16
-        with:
-          role-to-assume: ${{ secrets.AWS_DBFN_BUCKET_WRITE_ROLE }}
-          role-duration-seconds: 900 # 15 min; minimal duration possible
-          aws-region: us-east-1
-
-      - name: Upload DB Init Fn to AWS S3
-        run: aws s3 cp target/db-init.zip s3://lrsql-dbfn/db-init-${{ steps.version.outputs.version }}.zip
```
cliffcaseyyet, line 172: only thing that hits me is its weird that the bucket itself is not a var.

doc/aws.md
```diff
@@ -26,6 +26,23 @@ Note that if you do not use Route53 DNS you MUST provide one or more allowed COR
 
 In these templates the Load Balancer expects to provide access to the LRS via HTTPS/443. You will need to either acquire a free Amazon Certificate Manager cert (highly recommended) or import your own cert from another CA into ACM for use in the deployment.
 
+#### S3 Bucket (Non-US Regions Only)
+
+SQL LRS requires a Lambda function to create the application-level database user. For the following regions the code for this function is hosted by Yet Analytics and provided to the template automatically:
```
cliffcaseyyet, line 31: Maybe make a note to leave the DBInitFnBucketOverride /Key param blank.

.github/workflows/build.yml
milt: The bucket itself is not a var because then the automatic selection of region wouldn't work and US users would have another step. Note that the actual used bucket name contains `${AWS::Region}`

doc/aws.md
milt: Look further down in the step-by-step params, it does just that.

cliffcaseyyet (approved):

PR #291, opened 2023-03-30
SQL-181 RDS template updates

milt:
* Add `DBInstancePublic` boolean parameter, if `true` will make DB instances publicly accessible
* Add `DBInstanceReplica` boolean parameter, if `true` will add a 2nd DB instance

Also sets the default engine to `14.3` and makes the default instance `db.r5.large` which is a working combo.

### DIFF
--- a/dev-resources/template/1_db.yml
+++ b/dev-resources/template/1_db.yml
@@ -32,6 +32,20 @@ Parameters:
     Description: For provisioned aurora, the instance class to use
     Type: String
     Default: "db.r5.large"
+  DBInstanceReplica:
+    Description: Whether or not to create a second DB instance for use as a read-replica
+    Type: String
+    AllowedValues:
+      - true
+      - false
+    Default: false
+  DBInstancePublic:
+    Description: Whether or not to allow public access to DB instances
+    Type: String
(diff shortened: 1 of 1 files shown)


dev-resources/template/1_db.yml
```diff
@@ -24,14 +24,28 @@ Parameters:
...
-    Default: "db.r4.large"
+    Default: "db.r5.large"
+  DBInstanceReplica:
+    Description: Whether or not to create a second DB instance for use as a read-replica
+    Type: String
+    AllowedValues:
+      - true
+      - false
+    Default: false
+  DBInstancePublic:
```
cliffcaseyyet, line 42: I _would_ say to make the public align with the read replica only, or let them be set individually, but considering that you cant actually specify which instance is which i suppose its moot.

cliffcaseyyet (approved):

dev-resources/template/1_db.yml
milt: Yep, this is exactly why the public applies to both, only Jeff knows which will be which.

milt: exactly

PR #292, opened 2023-03-30
SQL-182 Set NoEcho on DefaultAdminPass to hide it

milt:
[SQL-182] from prying eyes

[SQL-182]: https://yet.atlassian.net/browse/SQL-182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cliffcaseyyet (commented):
this is a nice quick fix, but is it that much more to just source it from a securestring? I can't remember if that's one of the things that securestring makes into a huge pain in the template.

PR #295, opened 2023-04-07
SQL-187 Feature: Update Local Admin Password

milt:
[SQL-187] Allow updates to local admin passwords. Admin must provide old password to update.

- [X] Backend
- [X] API
- [X] Frontend - [PR here](https://github.com/yetanalytics/lrs-admin-ui/pull/43)

[SQL-187]: https://yet.atlassian.net/browse/SQL-187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/db/h2/lrsql/h2/sql/query.sql
+++ b/src/db/h2/lrsql/h2/sql/query.sql
@@ -243,9 +243,10 @@ WHERE activity_iri = :activity-iri
 -- :name query-account
 -- :command :query
 -- :result :one
--- :doc Given an account `username`, return the ID and the hashed password, which can be used to verify the account.
+-- :doc Given an account `username` or `account-id`, return the ID and the hashed password, which can be used to verify the account.
 SELECT id, passhash FROM admin_account
-WHERE username = :username
+--~ (when (:username params)   "WHERE username = :username")
+--~ (when (:account-id params) "WHERE id = :account-id")
 
 -- :name query-account-oidc
 -- :command :query
--- a/src/main/lrsql/admin/interceptors/account.clj
+++ b/src/main/lrsql/admin/interceptors/account.clj
@@ -8,20 +8,18 @@
             [lrsql.util :as u]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;; Validation Helpers
+;; Validation Interceptors
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(defn- validation-interceptor
-  "Shared helper for simple validation interceptors."
-  [interceptor-name
-   validate-spec
-   param-keys]
+(def validate-params
+  "Validate that the JSON params contain the params `username` and `password`
+   for login and create."
   (interceptor
-   {:name interceptor-name
+   {:name ::validate-params
     :enter
     (fn validate-params [ctx]
       (let [params (get-in ctx [:request :json-params])]
-        (if-some [err (s/explain-data validate-spec params)]
+        (if-some [err (s/explain-data ads/admin-params-spec params)]
           ;; Invalid parameters - Bad Request
           (assoc (chain/terminate ctx)
                  :response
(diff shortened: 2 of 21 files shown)


src/main/lrsql/admin/routes.clj
```diff
@@ -34,6 +34,13 @@
                                          ji/validate-jwt-account
                                          ai/create-admin)
      :route-name :lrsql.admin.account/create]
+    ;; Update account password
+    ["/admin/account/password"
+     :put (conj common-interceptors
```
kelvinqian00, line 39: PUT is "replace" while POST is "create" right?

milt: Usually, though in practice different systems do different things.

kelvinqian00: Yeah it seems lrsql follows convention (see create vs update API keys)

src/main/lrsql/admin/interceptors/account.clj
```diff
@@ -121,6 +144,39 @@
...
+                 {:status 404
+                  :body   {:error (format "The account username \"%s\" does not exist!"
+                                          username)}})
+
+          ;; The old password is not correct.
+          (= :lrsql.admin/invalid-password-error result)
+          (assoc (chain/terminate ctx)
+                 :response
+                 {:status 401
+                  :body   {:error "Invalid Account Credentials"}}))))}))
```
kelvinqian00, line 178: Actually one thing @cliffcaseyyet always emphasized when I worked on security implementations is that we should revel as few details as possible when returning security errors. For example, for `authenticate-admin` it would return a 401 error for both error results with a generic error message, and it seems like this should do the same.

kelvinqian00: Though on second look it might be okay to leave them separate here since you'd presumably _already_ be logged in with the username and (old) password. Which means that these should happen very rarely if ever, e.g. if another user concurrently deletes the account, but on those off-chance we'll still need these error cases.

cliffcaseyyet: To be fair in this case though, they are ALREADY authenticated. I usually advocate that at an authentication border.

milt:
As mentioned in slack, need to check against the account ID in the JWT. While I'm at it will probably amend the backend to just use the account ID instead of username

src/main/lrsql/spec/admin.clj
```diff
@@ -57,17 +62,42 @@
...
            (s/keys :req-un [:lrsql.spec.admin.input/oidc-enabled?])))
 
+(def update-admin-password-params-spec
+  (s/and
+   (s/keys :req-un [::old-password
+                    ::new-password])
+   (fn new-pass-noteq-old-pass
+     [{:keys [old-password
+              new-password]}]
+     (not= old-password new-password))))
```
kelvinqian00, line 95: Probably move this up to be with `admin-params-spec` and `admin-delete-params-spec` to avoid confusion between REST param specs and SQL query input specs.

src/db/h2/lrsql/h2/sql/query.sql
```diff
@@ -243,9 +243,10 @@ WHERE activity_iri = :activity-iri
 -- :name query-account
 -- :command :query
 -- :result :one
--- :doc Given an account `username`, return the ID and the hashed password, which can be used to verify the account.
+-- :doc Given an account `username` or `account-id`, return the ID and the hashed password, which can be used to verify the account.
 SELECT id, passhash FROM admin_account
-WHERE username = :username
+--~ (when (:username params)   "WHERE username = :username")
+--~ (when (:account-id params) "WHERE id = :account-id")
```
kelvinqian00, line 249: And when both are present, it performs an double check. 😈

src/main/lrsql/spec/admin.clj
milt: I moved it [SQL-187 move params spec to be with its siblings](https://github.com/yetanalytics/lrsql/pull/295/commits/5aaaea5eb702f40b9dcb56ac8f226c5997c3345d)

As for the other bit I think it is fine as part of the spec, but I'll add language to the form as you describe.

src/db/h2/lrsql/h2/sql/query.sql
milt: Is there a syntax for an OR between the two params? If so I'll use it, otherwise I don't think it matters too much.

src/main/lrsql/spec/admin.clj
milt: See https://github.com/yetanalytics/lrs-admin-ui/pull/43/commits/a993ea36f874b58f26c042b656d60f03f7dc07e4

src/db/h2/lrsql/h2/sql/query.sql
milt: Yeah that was my thinking too. However from outside there isn't really a way to even get to that point, since the params validation and extraction uses `select-keys` on the body params

cliffcaseyyet (approved):

src/db/h2/lrsql/h2/sql/query.sql
kelvinqian00: To be clear I don't think this is necessarily a problem (especially since we don't even do any doubled-up queries from the app), it was just an interesting observation.

kelvinqian00 (approved):

PR #298, opened 2023-05-09
[sql-189] Store timestamp and stored as normalized columns on xapi_statement

cliffcaseyyet:
- [x] Add column and migrate old data via DDL on startup. It should add a "timestamp" column in sqlite but thats just a string really. 
- It DOES perform TZ normalization of the timestamp of past data, which may be user-input offset
- It cannot produce a normalized timestamp above nanosecond precision so in the migration script for sqlite the zeroes are wedged in. In Postgres it sticks to 6 decimal places universally.

- [x] For all new records it normalizes both timestamp and stored exactly the same way, and modifies the existing util fn/macro to allow for offset inputs in timestamp which has not been needed before.


src/db/sqlite/lrsql/sqlite/sql/query.sql
```diff
@@ -320,10 +320,10 @@ GROUP BY platform
 -- :command :query
 -- :result :many
 -- :doc Return counts of statements by time unit for a given range.
-SELECT substr(json_extract(payload, "$.stored"), 1, :unit-for) stored,
+SELECT substr(json_extract(payload, "$.stored"), 1, :unit-for) AS stored_time,
```
cliffcaseyyet, line 323: conflict reduction

src/main/lrsql/util.clj
```diff
@@ -70,9 +74,13 @@
...
 
 (defn str->time
-  "Parse a string into a java.util.Instant timestamp."
+  "Parse an ISO 8601 timestamp string into a java.util.Instant timestamp. The
+  two parse fns are to support the Z and the +00:00 offset timestamp formats"
   [ts-str]
-  (wrap-parse-fn java-time/instant "timestamp" ts-str))
+  (wrap-parse-fn java-time/instant "timestamp" ts-str
+                 :retry-parse-fn #(-> %
+                                      java-time/offset-date-time
```
cliffcaseyyet, line 82: this extra step is unnecessary in newer versions of Java. This is basically just for 11 support where java.time/instant can't directly parse offsets like `+01:00` etc.

src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -280,3 +280,40 @@ CREATE TYPE scope_enum AS ENUM (
...
+
+-- :name alter-xapi-statement-add-timestamp!
+-- :command :execute
+-- :doc Add `xapi_statement.timestamp` to allow easier timestamp access.
+ALTER TABLE xapi_statement ADD COLUMN timestamp TIMESTAMPTZ
+
+-- :name migrate-xapi-statement-timestamps!
+-- :command :execute
+-- :doc Backfill `xapi_statement.timestamp` with the values from the payload
+UPDATE xapi_statement SET timestamp = (payload->>'timestamp')::timestamptz WHERE timestamp IS NULL;
```
kelvinqian00, line 301: Why the separate update functions? If you look at `alter-scope-enum-type!`, you'll see that multiple SQL updates are being performed in the same function.

cliffcaseyyet: Yeah, in general im a big fan of conditional transactions where available but in this case because of the architecture I like the isolation and the parity with sqlite commands. Since it's not something that runs all the time, really only once per instance, it doesn't affect performance or anything so i kinda dig it.

src/main/lrsql/util.clj
kelvinqian00: You should add that as a code comment then.

src/main/lrsql/spec/statement.clj
```diff
@@ -78,6 +79,7 @@
 ;; - verb_iri:         STRING NOT NULL
 ;; - is_voided:        BOOLEAN NOT NULL DEFAULT FALSE
 ;; - payload:          JSON NOT NULL
+;; TODO: timestamp and stored insts
```
kelvinqian00, line 82: You left a TODO here

cliffcaseyyet: eep, good call.

src/db/postgres/lrsql/postgres/sql/ddl.sql
kelvinqian00: > and the parity with sqlite commands

That makes sense, but I would like to point out that this ship had already sailed a long time ago with the OIDC issuer and passhash updates.

src/main/lrsql/util.clj
kelvinqian00: While we're in this code block, I would've done this differently by implementing a `str->time*` function that looks like this:
```clojure
(defn- str->time* [ts-str]
  (try (java-time/instant ts-str)
       (catch IllegalArgumentException _ ; or whatever java-time/instant thows
         (-> ts-str java-time/offset-date-time java-time/instant))))
```
instead of affecting the entire `wrap-parse-fn` macro.

src/db/postgres/lrsql/postgres/sql/ddl.sql
cliffcaseyyet: Yeah that's fine I'm not aiming to reinvent the wheel i just liked it in this case

src/main/lrsql/util.clj
cliffcaseyyet: Yes but I kinda loved the idea of one day reusing the macro for something ELSE that needed an alternative exception-triggered backup parser. It was more an experiment. Realized fully it could be a *set* of converters to try before dying but i didnt want to waste more time on it.

src/main/lrsql/spec/statement.clj
kelvinqian00: I see that you removed the TODO, but it's still missing the `stored` and `timestamp` fields

cliffcaseyyet: no, it's not.

cliffcaseyyet: oooh you mean the comments. oy.

src/main/lrsql/util.clj
kelvinqian00: In my view the updated macro is in an awkward spot right now:
- As you alluded, ideally this should support multiple possible backup functions (e.g. if plan B fails move onto plan C)
- If an exception happened on the backup function, it would not get caught (this is more pertinent in the short term)

cliffcaseyyet: Yes, i think it would, because it's recursively calling the same macro.

kelvinqian00: Oh good point. You should add a test for that though to make sure that that's what _actually_ happens.

cliffcaseyyet: Yeah ok I thought about a fail case, that's rational.

src/main/lrsql/spec/statement.clj
cliffcaseyyet: 9b8ed91c0da4c20b6b76c950bd700952a4fb47ac

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
```diff
@@ -328,3 +328,41 @@ SET sql = 'CREATE TABLE credential_to_scope (
...
+-- :name query-xapi-statement-timestamp-exists
+-- :command :query
+-- :result :one
+-- :doc Query to see if `xapi_statement.timestamp` exists.
+SELECT 1 FROM pragma_table_info('xapi_statement') WHERE name = 'timestamp'
+
+-- :name alter-xapi-statement-add-timestamp!
+-- :command :execute
+-- :doc Add `xapi_statement.timestamp` to allow easier timestamp access.
+ALTER TABLE xapi_statement ADD COLUMN timestamp TIMESTAMP
```
kelvinqian00, line 343: I overlooked this but I didn't realize that SQLite had a `TIMESTAMP` type keyword. If you looked at the `last_modified` property elsewhere in the DDL you'll see I'll just have their type as `TEXT`.

src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -280,3 +280,40 @@ CREATE TYPE scope_enum AS ENUM (
...
+-- :name query-xapi-statement-timestamp-exists
+-- :command :query
+-- :result :one
+-- :doc Query to see if `xapi_statement.timestamp` exists.
+SELECT 1 FROM information_schema.columns WHERE table_name = 'xapi_statement' AND column_name = 'timestamp';
+
+-- :name alter-xapi-statement-add-timestamp!
+-- :command :execute
+-- :doc Add `xapi_statement.timestamp` to allow easier timestamp access.
+ALTER TABLE xapi_statement ADD COLUMN timestamp TIMESTAMPTZ
```
kelvinqian00, line 296: So statement `stored` and `timestamp` are `TIMESTAMPZ`, but `last_modified` in other tables are just plain `TIMESTAMP`. That can get confusing.

src/main/lrsql/spec/statement.clj
kelvinqian00: `- stored: STORED NOT NULL` - don't you mean `TIMESTAMP NOT NULL`?

cliffcaseyyet: hah, brain fart i knew that

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
cliffcaseyyet: its a do-nothing alias, but in this case a useful one for inspection purposes

src/db/postgres/lrsql/postgres/sql/ddl.sql
cliffcaseyyet: Well as ive explained, you need TZ to actually represent normalized zulu

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
kelvinqian00: What would happen if we changed the `TEXT` type in those instances with `TIMESTAMP` in the SQLite DDL?

src/main/lrsql/spec/statement.clj
cliffcaseyyet: fixed

src/main/lrsql/util.clj
```diff
@@ -25,15 +25,19 @@
...
-                          {:type      ::parse-failure
-                           :data      ~data
-                           :data-type ~data-type})))))
+          ~(if (some? retry-parse-fn)
+             `(wrap-parse-fn ~retry-parse-fn ~data-type ~data)
+             `(throw (ex-info (format "Cannot parse nil or invalid %s"
+                                      ~data-type)
+                              {:type      ::parse-failure
+                               :data      ~data
+                               :data-type ~data-type}))))))
```
milt, line 40: Ahh, this looks familiar

cliffcaseyyet: Who's to say from where inspiration springs.

milt (approved):

src/db/postgres/lrsql/postgres/sql/ddl.sql
kelvinqian00: Discussed on Slack - such a concern would need to be addressed in a future PR.

kelvinqian00 (approved):

PR #300, opened 2023-05-12
Timestamp updates

cliffcaseyyet:
Updating last_modified on document tables to use timestamp/timstamptz. Migration of old timestamps in place with zone awareness


src/db/sqlite/lrsql/sqlite/sql/ddl.sql
```diff
@@ -329,7 +329,7 @@ SET sql = 'CREATE TABLE credential_to_scope (
 )'
 WHERE type = 'table' AND name = 'credential_to_scope'
 
-/* Migration 2022-05-08-00 - Add timestamp to xapi_statement */
+/* Migration 2023-05-08-00 - Add timestamp to xapi_statement */
```
kelvinqian00, line 332: It's the "it's still the previous year!" feeling isn't it?

cliffcaseyyet: yep

src/db/postgres/lrsql/postgres/record.clj
```diff
@@ -59,7 +60,13 @@
       (migrate-xapi-statement-timestamps! tx))
     (when-not (some? (query-xapi-statement-stored-exists tx))
       (alter-xapi-statement-add-stored! tx)
-      (migrate-xapi-statement-stored-times! tx)))
+      (migrate-xapi-statement-stored-times! tx))
+    (when-not (some? (query-state-document-last-modified-is-timestamptz tx))
```
kelvinqian00, line 64: I was going ask why you needed this separate query until I saw that you're using three different functions to do the PG migration. Even then I think you could've just added the conditions directly in the commands, but I guess this way is clearer, especially if you're just looking at the Clojure code.

cliffcaseyyet: yeah i mean it just skips the whole chunk in the future.

src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -317,3 +317,29 @@ ALTER TABLE xapi_statement ADD COLUMN stored TIMESTAMPTZ
...
 -- :doc Backfill `xapi_statement.stored` with the values from the payload
 UPDATE xapi_statement SET stored = (payload->>'stored')::timestamptz WHERE stored IS NULL;
+
+/* Migration 2023-05-11-00 - Convert timestamps for consistency */
+
+-- :name query-state-document-last-modified-is-timestamptz
+-- :command :query
+-- :result :one
+-- :doc Query to see if `state_document.last_modified` is a timestamp.
+SELECT * FROM information_schema.columns WHERE table_name = 'state_document' AND column_name = 'last_modified' and data_type = 'timestamp with time zone';
```
kelvinqian00, line 327: Probably should break this query up into separate lines. Also why is the last `and` lowercase?

cliffcaseyyet: 62ce52092f060df54af2fa6469fb8f358c21de15

src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -89,6 +89,19 @@
...
+      (migrate-timestamps-state-03! tx)
+      (migrate-timestamps-state-04! tx)
+      (migrate-timestamps-agent-profile-01! tx)
+      (migrate-timestamps-agent-profile-02! tx)
+      (migrate-timestamps-agent-profile-03! tx)
+      (migrate-timestamps-agent-profile-04! tx)
+      (migrate-timestamps-activity-profile-01! tx)
+      (migrate-timestamps-activity-profile-02! tx)
+      (migrate-timestamps-activity-profile-03! tx)
+      (migrate-timestamps-activity-profile-04! tx))
```
kelvinqian00, line 104: As you said this looks absolutely disgusting, but given that it's just a one-time migration and the whole thing is conceptually simple it's an okay workaround. Really wish I knew of a way to do multi-statement commands in SQLite...

src/db/postgres/lrsql/postgres/record.clj
```diff
@@ -59,7 +60,13 @@
...
+      (let [tz-input
+            {:tz-id (str "'" (.getID (.getTimeZone (Calendar/getInstance))) "'")}]
```
kelvinqian00, line 66: This should definitely be extracted out into its own function. At the very least the ugly Java interop can go into the main `util` namespace. You could also in principle put the `:tz-id` wrapping in a separate `input/` namespace, though that would probably be overkill.

cliffcaseyyet: actually i like that idea. keep the muck in input

cliffcaseyyet: oooo wait a minute though this is engine-specific

cliffcaseyyet: https://github.com/yetanalytics/lrsql/pull/300/commits/6dd5400dbd77ecaa5bd66617c1e9fce1f85fa653

src/main/lrsql/util.clj
```diff
@@ -14,7 +14,8 @@
   (:import [java.util UUID]
            [java.time Instant]
            [java.io StringReader PushbackReader ByteArrayOutputStream]
-           [java.nio.charset Charset]))
+           [java.nio.charset Charset]
+           [java.util Calendar]))
```
kelvinqian00, line 18: You can put `Calendar` alongside `UUID` since they're both being imported from `java.util`.

cliffcaseyyet: nice catch, missed that entirely. https://github.com/yetanalytics/lrsql/pull/300/commits/c22c56150325dbde64bdd46941cb1cddae7cfeeb

kelvinqian00 (approved):

milt (approved):

PR #305, opened 2023-07-10
JSONB option and support

cliffcaseyyet:
Added JSONB as an option to allow for analytics-tuned postgres performance


doc/env_vars.md
```diff
@@ -80,6 +80,14 @@ You may have noted that some options are not available:
 - `healthCheckRegistry` cannot easily report via JMX, and most of its information should be covered by `metricRegistry` anyways.
 - `threadFactory` and `scheduledExecutor` are Java instances that should only be used in specific execution environments.
 
+### Tuning
+
+The following options are used for advanced database performance tuning and may affect DB schema and performance. Use with caution.
+
+| Env Var              | Config        | Description                         | Default |
+| -------------------  | ------------- | ----------------------------------- | ------- |
+| `LRSQL_ENABLE_JSONB` | `enableJsonb` | (Postgres Only) This option switches the `JSON` fields (including `xapi_statment.payload`) to `JSONB` datatype. Enabling this option increases the speed of some external JSON-based queries such as from a BI or reporting platform, but decreases statement write throughput. Be advised that changing this setting on a large existing database may result in a very long next startup time (possibly minutes) as all previous data gets converted. | `false` |
```
milt, line 89: Is there like a tx timeout param we could mention here for large DBs that would help?

milt (approved):
Useful!

resources/lrsql/config/config.edn
```diff
@@ -37,6 +37,17 @@
    :prod-sqlite     #include "prod/default/connection.edn"
    :prod-sqlite-mem #include "prod/default/connection.edn"
    :prod-postgres   #include "prod/postgres/connection.edn"}
+ :tuning
+ #profile
```
kelvinqian00, line 41: We discussed this before: while I, as a developer, think that any tuning config should be subordinate to overall DB config, having tuning config be top-level may be easier to grasp for end users. Is that the reasoning behind putting tuning config at the top?

src/db/postgres/lrsql/postgres/record.clj
```diff
@@ -50,7 +50,7 @@
     (create-admin-account-table! tx)
     (create-credential-table! tx)
     (create-credential-to-scope-table! tx))
-  (-update-all! [_ tx]
+  (-update-all! [{{{jsonb? :enable-jsonb} :config} :tuning} tx]
```
kelvinqian00, line 53: I believe that you can use `tuning` directly as a record field, since Component will add it upon start. Definitely try that to see if it works.

resources/lrsql/config/config.edn
cliffcaseyyet: So I'm trying to look at the category more broadly. There may in fact be CODE changes in the future that adjust the performance profile of an LRS for different use cases.

src/main/lrsql/system.clj
```diff
@@ -2,6 +2,7 @@
   (:require [com.stuartsierra.component :as component]
             [lrsql.system.database :as db]
```
kelvinqian00, line 3: Now that we have a separate DB-centric config in tuning, having it outside the `system.database` namespace can get confusing. Either move the tuning record to `system.database` or rename it to `system.connection`, since that is the record that namespace is currently focused on.

cliffcaseyyet: see above.

src/main/lrsql/system.clj
kelvinqian00, line 1: I'm surprised that you didn't refactor how configuration is introduced into the records, despite the amount of confusion it caused you. (Also it's annoying that you can't add reviews to unedited lines on GitHub.)

src/db/postgres/lrsql/postgres/record.clj
kelvinqian00:
```clojure
(defrecord PostgresBackedn [tuning]
  ...
  (-update-all! [_ tx]
    ...
    (if (-> tuning :config :enable-jsonb)
      (migrate-to-jsonb! tx)
      (migrate-to-json! tx)))
```

resources/lrsql/config/config.edn
kelvinqian00: Ah, so code changes that have nothing to do with adjusting the DDL or the queries?

cliffcaseyyet: I could see it. Batching strategies, timeout stuff, etc. oooo i just realized that txn-retries might be a good candidate to be moved here. What do you think?

cliffcaseyyet: Also it should be noted that milt is building some pretty heavy duty stuff that's gonna get tacked on and may itself have some tuning properties.

src/db/postgres/lrsql/postgres/record.clj
cliffcaseyyet: word, that tracks, i'll see if it plays.

src/main/lrsql/system.clj
cliffcaseyyet: Yeah i didn't feel like pulling it inside out. I am down with writing a comment though.

kelvinqian00: Sure, whatever you think would make you not confused the next time you look at this.

cliffcaseyyet: https://github.com/yetanalytics/lrsql/pull/305/commits/f41d5a334efb5cdec3e929d06f54e2e72b5a6028

src/db/postgres/lrsql/postgres/record.clj
cliffcaseyyet: https://github.com/yetanalytics/lrsql/pull/305/commits/f41d5a334efb5cdec3e929d06f54e2e72b5a6028

kelvinqian00 (approved):

PR #307, opened 2023-07-21
SQL-197 Reaction Persistence

milt:
[SQL-197] [SQL-198] Add database persistence for Reactions.

Creates backend methods to insert, read, update and soft-delete reactions.

[SQL-197]: https://yet.atlassian.net/browse/SQL-197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SQL-198]: https://yet.atlassian.net/browse/SQL-198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/db/sqlite/lrsql/sqlite/sql/ddl.sql
```diff
@@ -426,3 +426,33 @@ ALTER TABLE activity_profile_document DROP COLUMN last_modified;
...
+
+
+/* Migration 2023-07-21-00 - Add Reaction Table */
+
+-- :name create-reaction-table!
+-- :command :execute
+-- :doc Create the `reaction` table if it does not yet exist.
+CREATE TABLE IF NOT EXISTS reaction (
+  id         TEXT NOT NULL PRIMARY KEY, -- uuid
+  ruleset    BLOB NOT NULL,             -- serialized reaction spec
```
cliffcaseyyet, line 438: BLOB because transit or something?

src/db/sqlite/lrsql/sqlite/sql/delete.sql
```diff
@@ -59,3 +59,13 @@ DELETE FROM credential_to_scope
 WHERE api_key = :api-key
 AND secret_key = :secret-key
 AND scope = :scope
+
+-- :name delete-reaction!
+-- :command :execute
+-- :result :affected
+-- :doc (Soft) delete a reaction.
```
cliffcaseyyet, line 66: good call on soft delete for statement FK

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
cliffcaseyyet: Also I presume this is where templates will be added or are we doing them in another table with a FK?

cliffcaseyyet (commented):
Very straightforward.

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
milt: BLOB because that's what the other JSON stuff used. I think the templates should just go in as part of the same structure.

src/db/sqlite/lrsql/sqlite/sql/delete.sql
milt: Yeah I think it makes sense, even if it is the only soft delete in the system (not counting voids which are kind of one)

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
cliffcaseyyet: ok sounds good. I forgot we didnt use a json type, since it doesnt buy us anything at all

cliffcaseyyet (approved):

invaliduser (approved):

PR #310, opened 2023-07-26
Fixed typo in protocol method

milt:
It was in a protocol declaration so it didn't really mess anything up, but still.


kelvinqian00 (approved):
Praytell there are not any other typos in the code

PR #313, opened 2023-07-28
SQL-193 Statement Generation FN

invaliduser:
[SQL-193]

[SQL-193]: https://yet.atlassian.net/browse/SQL-193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/lrsql/util/reaction.clj
+++ b/src/main/lrsql/util/reaction.clj
@@ -1,6 +1,7 @@
 (ns lrsql.util.reaction
   "Utilities to support reactions."
   (:require [clojure.spec.alpha :as s]
+            [clojure.walk :as walk]
             [lrsql.spec.reaction :as rs]
             [lrsql.spec.statement :as ss]
             [xapi-schema.spec :as xs]
@@ -101,3 +102,23 @@
   [ruleset-bytes]
   (with-open [r (io/reader ruleset-bytes)]
     (cjson/parse-stream r (partial keyword nil))))
+
+(s/fdef generate-statement
+  :args (s/cat :cond-map (s/map-of ::rs/condition-name
+                         ::xs/statement)
+               :template ::xs/any-json)
+  :ret ::xs/statement)
+
+(defn generate-statement [cond->statement template]
+  (walk/postwalk #(if (and (map? %)
+                           (= (key (first %)) "$templatePath"))
+                    (let [input-path (val (first %))
+                          path (update input-path 0 keyword)
+                          result (get-in cond->statement path :not-found)]
+                      (case result
+                        :not-found (throw (ex-info (str "No value found at " input-path)
+                                                   {:path input-path
+                                                    :type ::invalid-path}))
+                        result))
+                    %)
+                 template))
--- a/src/test/lrsql/util/reaction_test.clj
+++ b/src/test/lrsql/util/reaction_test.clj
@@ -90,3 +90,76 @@
            (-> simple-reaction-ruleset
                r/serialize-ruleset
                r/deserialize-ruleset)))))
+
+
+(def fake-cond-map
+  {:completed_a {"id"     "6fbd600f-d17c-4c74-801a-2ec2e53231f7"
+                 "actor"  {"mbox"       "mailto:bob@example.com"
+                           "madeUpAttr" "8"}
+                 "verb"   {"id" "https://example.com/verbs/completed"}
+                 "object" {"id"         "https://example.com/activities/a"
+                           "objectType" "Activity"}
+                 "result" [{"success" true} {"success" false}]}
+   :completed_b {"id"     "c51d1628-ae4a-449f-8d8d-13d57207f468"
+                 "actor"  {"mbox" "mailto:bob@example.com"}
+                 "verb"   {"id" "https://example.com/verbs/completed"}
+                 "object" {"id"         "https://example.com/activities/b"
+                           "objectType" "Activity"}
+                 "result" [{"success" false} {"success" true}]}})
+
+(deftest statement-generation-test
+  (testing "Basic statement generation"
+    (is (= (r/generate-statement
+            fake-cond-map
+            {"actor" {"mbox" {"$templatePath" ["completed_a" "actor" "mbox"]}}
+             "verb"  "did"
+             "object"  "this"})
+           {"actor" {"mbox" "mailto:bob@example.com"}
+            "verb"  "did"
+            "object"  "this"})))
+
+  (testing "Inclusion of multiple attributes"
+    (is (= (r/generate-statement
+            fake-cond-map
+            {"actor" {"mbox" "mailto:bob@example.com"
+                      "madeUpAttr" {"$templatePath" ["completed_a" "actor" "madeUpAttr"]}}
+             "verb"  "did"
+             "object"  "this"})
+           {"actor" {"mbox" "mailto:bob@example.com"
+                     "madeUpAttr" "8"}
+            "verb"  "did"
+            "object"  "this"})))
+  (testing "non-string contents"
+    (is (= (r/generate-statement
+            fake-cond-map
+            {"verb"   {"$templatePath" ["completed_b" "verb"]}})
+           {"verb"  {"id" "https://example.com/verbs/completed"}})))
+
+  (testing "generate statements patched together from different conditions"
+    (is (= (r/generate-statement
+            fake-cond-map
+            {"actor"  {"$templatePath" ["completed_a" "actor"]}
+             "verb"   "did"
+             "object" {"$templatePath" ["completed_b" "object"]}})
+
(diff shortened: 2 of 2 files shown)


src/test/lrsql/ops/query/generate_test.clj
```diff
@@ -0,0 +1,72 @@
+(ns lrsql.ops.query.generate-test
+  (:require [clojure.test :refer [deftest testing is]]
+            [lrsql.ops.query.generate :as generate]))
+
+(def fake-cond-map
+  {"completed_a" {"id"     "6fbd600f-d17c-4c74-801a-2ec2e53231f7"
```
milt, line 6: The outermost keys in the cond-map returned from SQL will be kehywords, not strings. The rest is correct though.

src/main/lrsql/ops/query/generate.clj
```diff
@@ -0,0 +1,16 @@
+(ns lrsql.ops.query.generate
```
milt, line 1: The ops/(query|command) namespaces are generally for things that do queries, not pure data stuff like this. I'd suggest moving this somewhere like `lrsql.util.reaction`

src/main/lrsql/ops/query/generate.clj
```diff
@@ -0,0 +1,16 @@
...
+  (:require [clojure.spec.alpha :as s]
+            [clojure.walk :as walk]
+            [lrsql.spec.reaction :as react-spec]))
+
+(s/fdef generate
+  :args (s/cat :cond-map (s/map-of :react-spec/condition-name
+                         :react-spec/val)
+               :statement map?))
+
+(defn generate [cond->statement template]
```
milt, line 11: Don't know if `generate` is a great name for this, it's a little too general. Maybe something like generate-statement if you move it to that reactions util NS

src/main/lrsql/ops/query/generate.clj
```diff
@@ -0,0 +1,16 @@
+(ns lrsql.ops.query.generate
+  (:require [clojure.spec.alpha :as s]
+            [clojure.walk :as walk]
+            [lrsql.spec.reaction :as react-spec]))
+
+(s/fdef generate
+  :args (s/cat :cond-map (s/map-of :react-spec/condition-name
```
milt, line 7: when you use an alias keyword like this, you need double dots, like `::react-spec/condition-name`

src/test/lrsql/ops/query/generate_test.clj
invaliduser: changed

src/main/lrsql/ops/query/generate.clj
invaliduser: moved

src/main/lrsql/ops/query/generate.clj
invaliduser: renamed to `generate-statement`

src/main/lrsql/ops/query/generate.clj
invaliduser: fixed

src/test/lrsql/ops/query/generate_test.clj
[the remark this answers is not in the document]
invaliduser: I changed it to throw an exception (that provides the path) in this case.

src/main/lrsql/util/reaction.clj
```diff
@@ -101,3 +102,19 @@
   [ruleset-bytes]
   (with-open [r (io/reader ruleset-bytes)]
     (cjson/parse-stream r (partial keyword nil))))
+
+(s/fdef generate-statement
+  :args (s/cat :cond-map (s/map-of ::rs/condition-name
+                         ::rs/val)
```
milt, line 108: `::rs/val` is an open-ended spec covering any JSON. In this case you actually have a good spec to draw on, `:xapi-schema.spec/statement`, as anything coming out of that query should be a valid xAPI statement

src/main/lrsql/util/reaction.clj
```diff
@@ -101,3 +102,19 @@
...
+               :statement map?))
```
milt, line 109: In this case, we're really talking about a template, so the `s/cat` key should probably be `:template` (I see it is named that down in the function so this should match). As for the spec we _can't_ rely on it being a valid xAPI statement like we can above, because of the additional structures for paths so `map?` is OK, but we can be a *little* more specific by using `:xapi-schema.spec/any-json`.

milt: Also we're missing a `:ret` spec here, which would be `:xapi-schema.spec/statement` I believe.

src/test/lrsql/util/reaction_test.clj
```diff
@@ -90,3 +90,68 @@
...
+                  "verb"   {"id" "https://example.com/verbs/completed"}
+                  "object" {"id"         "https://example.com/activities/a"
+                            "objectType" "Activity"}
+                  "result" [{"success" true} {"success" false}]}
+   :completed_b   {"id"     "c51d1628-ae4a-449f-8d8d-13d57207f468"
+                    "actor"  {"mbox" "mailto:bob@example.com"}
+                    "verb"   {"id" "https://example.com/verbs/completed"}
+                    "object" {"id"         "https://example.com/activities/b"
+                              "objectType" "Activity"}
+                    "result" [{"success" false} {"success" true}]}})
```
milt, line 108: Formatting is a little odd on these maps. If you have the Emacs `clojure-align` command, that will nicely format them for you with vertical alignment. Catch me on slack if you have trouble getting that set up.

src/main/lrsql/util/reaction.clj
```diff
@@ -101,3 +102,19 @@
...
+
+(defn generate-statement [cond->statement template]
+  (walk/postwalk #(if (and (map? %)
+                           (= (key (first %)) "$templatePath"))
+                    (let [path (val (first %))
+                          result (get-in cond->statement path :not-found)]
+                      (case result
+                        :not-found (throw (Exception. (str "No value found at " path)))
```
milt, line 117: Generally by convention we throw `clojure.lang.ExceptionInfo` errors, so we can tack on any relevant data. Check out `clojure.core/ex-info` for that. The convention is we pick a namespaced `:type` key to give them that is descriptive of the error (ex. `::invalid-path`), and then any other stuff like the path itself.

src/test/lrsql/util/reaction_test.clj
```diff
@@ -90,3 +90,68 @@
...
+
+(deftest statement-generation-test
+  (testing "Basic statement generation"
+    (is (= (r/generate-statement
+            fake-cond-map
+            {"actor" {"mbox" {"$templatePath" [:completed_a "actor" "mbox"]}}
```
milt, line 114: Since the input for this will come from JSON, we won't be able to mix keywords and strings in the path like this. Your function should probably handle the first segment being a key so that the path could just be `["completed_a" "actor" "mbox"]`

src/test/lrsql/util/reaction_test.clj
```diff
@@ -90,3 +90,76 @@
...
+            "b_first_result" false})))
+
+  (testing "when no value at path, throws appropriate exception"
+    (is (= (try (r/generate-statement
+                  fake-cond-map
+                  {"actor_birthday" {"$templatePath" ["completed_a" "actor" "birthday"]}})
+
+                (catch clojure.lang.ExceptionInfo e (ex-info "No value found at [:completed_a \"actor\" \"birthday\"]"
+                                                             {:path ["completed_a" "actor" "birthday"]
+                                                              :type ::invalid-path})))))))
```
milt, line 165: Check your parens on this `=`.

src/test/lrsql/util/reaction_test.clj
```diff
@@ -90,3 +90,76 @@
...
+             "b_first_result" {"$templatePath" ["completed_b" "result" 0 "success"]}})
+           {"a_first_result" true
+            "b_first_result" false})))
+
+  (testing "when no value at path, throws appropriate exception"
+    (is (= (try (r/generate-statement
+                  fake-cond-map
+                  {"actor_birthday" {"$templatePath" ["completed_a" "actor" "birthday"]}})
+
+                (catch clojure.lang.ExceptionInfo e (ex-info "No value found at [:completed_a \"actor\" \"birthday\"]"
```
milt, line 163: Two points of convention:
* We generally try to keep code to an 80 char horizontal max (this one is looking a little wide)
* We generally line break after the exception binding in `catch` block. Not something I even think about but all of our code thusfar has done it, lol.

milt (approved):
Looks great!

PR #315, opened 2023-08-07
SQL-200 Reaction TX Function & Errors

milt:
[SQL-200] Finish core reaction transaction function. Store errors when they happen and disable reactions with errors.

[SQL-200]: https://yet.atlassian.net/browse/SQL-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/db/sqlite/lrsql/sqlite/sql/ddl.sql
+++ b/src/db/sqlite/lrsql/sqlite/sql/ddl.sql
@@ -438,7 +438,8 @@ CREATE TABLE IF NOT EXISTS reaction (
   ruleset    BLOB NOT NULL,             -- serialized reaction spec
   created    TIMESTAMP NOT NULL,        -- timestamp
   modified   TIMESTAMP NOT NULL,        -- timestamp
-  active     INTEGER                    -- true/false/null - active/inactive/soft delete
+  active     INTEGER,                   -- true/false/null - active/inactive/soft delete
+  error      BLOB                       -- serialized error
 )
 
 -- :name query-xapi-statement-reaction-id-exists
--- a/src/main/lrsql/ops/query/reaction.clj
+++ b/src/main/lrsql/ops/query/reaction.clj
@@ -135,5 +135,8 @@
                            ;; Success Response
                            {:reaction-id reaction-id
                             :trigger-id  trigger-id
-                            :statement   statement}))))))))))
+                            :statement   (ru/add-reaction-metadata
+                                          statement
+                                          reaction-id
+                                          trigger-id)}))))))))))
        active-reactions))}))
--- a/src/main/lrsql/ops/query/reaction.clj
+++ b/src/main/lrsql/ops/query/reaction.clj
@@ -139,8 +139,8 @@
                                           new-statement
                                           reaction-id
                                           trigger-id)
-                            ;; Use a custom authority from the ruleset or use
+                            ;; Use a custom authority from the template or use
                             ;; the trigger statement's authority
-                            :authority   (get ruleset :authority
-                                              (get statement "authority"))}))))))))))
+                            :authority   (or (get new-statement "authority")
+                                             (get statement "authority"))}))))))))))
        active-reactions))}))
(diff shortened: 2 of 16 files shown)


src/db/sqlite/lrsql/sqlite/sql/ddl.sql
```diff
@@ -438,7 +438,8 @@ CREATE TABLE IF NOT EXISTS reaction (
   ruleset    BLOB NOT NULL,             -- serialized reaction spec
   created    TIMESTAMP NOT NULL,        -- timestamp
   modified   TIMESTAMP NOT NULL,        -- timestamp
-  active     INTEGER                    -- true/false/null - active/inactive/soft delete
+  active     INTEGER,                   -- true/false/null - active/inactive/soft delete
```
cliffcaseyyet, line 441: not sure how i feel about the trinary vs another field for deleted, I'm usually not a fan, but I guess I can't practically see an issue with it.

src/main/lrsql/system/lrs.clj
```diff
@@ -375,4 +376,37 @@
...
+             (:result
+              (react-q/query-statement-reactions
+               backend tx {:trigger-id statement-id}))))]
+      ;; Submit statements one at a time with varying authority
+      {:statement-ids
+       (reduce
+        (fn [acc {:keys [statement authority]}]
+          (into acc
+                (:statement-ids
+                 (lrsp/-store-statements
```
cliffcaseyyet, line 405: i dont see where it would loop from reactions, but actually i guess that flow is what you're working on now? Possibly triggered from store-statements?

cliffcaseyyet: Also I see what you mean about store-statements needing to repeat due to authority, but to be fair batch prep isnt ACTUALLY that efficient in lrsql. The db is still called n times.

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
milt: Yep, that is why we are soft-deleting. As far as the trinary goes, we used it on previous projects so I thought it was OK, but am happy to change if desired.

src/main/lrsql/ops/query/reaction.clj
```diff
@@ -45,76 +56,91 @@
...
+     (into
+      []
+      (mapcat
+       (fn [{:keys       [ruleset]
+             reaction-id :id}]
+         ;; Cycle Check
+         (if (contains? s-reactions reaction-id)
+           (do
+             (log/warnf
+              "Reaction %s found in statement %s history, ignoring!"
```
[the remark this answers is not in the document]
milt: It should be, yeah!

src/main/lrsql/system/lrs.clj
milt: > Possibly triggered from store-statements?

Yep, the current plan is to do a non-blocking submit to the queue (channel) from store-statements after everything is done, conditional on this feature being enabled ofc.

src/main/lrsql/util/reaction.clj
```diff
@@ -93,6 +93,16 @@
   (let [out-stream (ByteArrayOutputStream. 4096)]
     (.toByteArray ^ByteArrayOutputStream (write-json* out-stream ruleset))))
 
+(s/fdef serialize-error
+  :args (s/cat :error ::rs/error)
+  :ret bytes?)
+
+(defn serialize-error
```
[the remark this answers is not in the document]
milt: I was following what I saw the other write queries (like for statements) doing for JSON, which was a JSON-encoded byte array. I don't believe you can just pass the data and have it magically be JSON, but not sure as the backend does seem to auto-coerce _some_ stuff.

cliffcaseyyet (approved):

PR #321, opened 2023-08-18
SQL-196 Reaction Management API

milt:
[SQL-196] Expose CRUD actions for reactions via the API

### DIFF
--- a/src/main/lrsql/spec/reaction.clj
+++ b/src/main/lrsql/spec/reaction.clj
@@ -232,3 +232,19 @@
 
 (def error-reaction-ret-spec
   (s/keys :req-un [:lrsql.spec.reaction.error-reaction/result]))
+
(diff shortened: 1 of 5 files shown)


src/main/lrsql/spec/reaction.clj
```diff
@@ -232,3 +232,19 @@
...
+;; Params
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(def create-reaction-params-spec
+  (s/keys :req-un [::ruleset
+                   ::active]))
+
+(def update-reaction-params-spec
+  (s/keys :req-un [::reaction-id
+                   (or ::ruleset
```
cliffcaseyyet, line 246: Why is this an or? You can't update both?

milt: I believe this just requires that _one_ of them be present, though it will also let them both be.

cliffcaseyyet (approved):

PR #325, opened 2023-08-23
delete-actor

invaliduser:
additions to sqlite/record.clj, appropriate additions to delete.sql and query.sql, addition to backend/protocol.clj


src/db/sqlite/lrsql/sqlite/sql/query.sql
```diff
@@ -327,3 +327,59 @@ WHERE id > :since-id
   AND id <= :until-id
 GROUP BY stored_time
 ORDER BY stored_time ASC;
+
+-- :name delete-actor-st2actor!
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_actor where actor_ifi = :actor-ifi
```
kelvinqian00, line 334: Why do you have DELETE statements in a query.sql file? I'm assuming you copy-pasted this here without looking carefully at where you pasted.

src/db/sqlite/lrsql/sqlite/sql/delete.sql
```diff
@@ -59,3 +59,53 @@ DELETE FROM credential_to_scope
...
+DELETE FROM statment_to_activity sta 
+LEFT JOIN xapi_statement ON sta.statement_id=xapi_statement.statement_id 
+WHERE statement_to_activity.statement_id IN (:del-statement-ids)
+
+-- :name delete-actor-delete-reactions!
+-- :command :execute
+-- :result :affected   
+DELETE FROM reactions 
+LEFT JOIN xapi_statement ON reactions.id = xapi_statement.reaction_id 
+WHERE xapi_statement.statement_id IN (:del-statement-ids)
```
kelvinqian00, line 95: I'm not sure if we should implement cascades on reactions when they're not fully implemented yet. @milt what do you think?

kelvinqian00:
Remember that SQLite is only half of the SQL LRS backend; there's also the Postgres version that will need to be implemented.

src/db/sqlite/lrsql/sqlite/sql/query.sql
kelvinqian00: Actually since you mentioned it in the PR description, it seems like it was actually intentional? In that case query.sql should _only_ contain queries, and no DELETE statements or other mutations.

invaliduser: oh dear, let me fix that

kelvinqian00:
I also noticed that you did not use any instances of ON DELETE CASCADE. Which is fair, since that would require changes to the DDL which will require SQL migrations, and those are always inherently dodgy (at least moreso than just adding cascades manually like you have here).

src/db/sqlite/lrsql/sqlite/sql/delete.sql
cliffcaseyyet: No definitely not. Also, i cannot devise a way in which reactions would need to be deleted at all, even when thats merged in. This is looking at it backwards. Reaction id is just a tag on a statement, you can safely delete the statement and the data goes away.

src/db/sqlite/lrsql/sqlite/sql/delete.sql
```diff
@@ -59,3 +59,53 @@ DELETE FROM credential_to_scope
...
+
+-- :name delete-actor-st-desc-rels!
+-- :command :execute
+-- :result :affected   
+DELETE FROM statement_to_statement WHERE ancestor_id in (:del-statement-ids)
+
+-- :name delete-actor-st2activ!
+-- :command :execute
+-- :result :affected   
+DELETE FROM statment_to_activity sta
```
cliffcaseyyet, line 86: typo?

src/db/sqlite/lrsql/sqlite/sql/delete.sql
```diff
@@ -59,3 +59,53 @@ DELETE FROM credential_to_scope
...
+
+-- :name delete-actor-st2st!
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_statement WHERE descendant_id in (:del-statement-ids)
+
+-- :name delete-actor-st-parent-rels!
+-- :command :execute
+-- :result :affected   
+DELETE FROM statement_to_statement WHERE descendant_id in (:del-statement-ids)
```
cliffcaseyyet, line 76: this and the one above are identical queries.

src/db/sqlite/lrsql/sqlite/sql/query.sql
```diff
@@ -327,3 +327,10 @@ WHERE id > :since-id
   AND id <= :until-id
 GROUP BY stored_time
 ORDER BY stored_time ASC;
+
+-- :name query-statement-ids-by-actor
+-- :command :query
+-- :result :many
+SELECT x.id FROM xapi_statement x
+LEFT JOIN statement_to_actor sta ON x.statement_id=sta.statement_id
```
cliffcaseyyet, line 335: this is for sure not getting what you need.

src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -134,6 +134,17 @@
     (insert-statement-to-actor! tx input))
   (-update-actor! [_ tx input]
     (update-actor! tx input))
+  (-delete-actor! [_ tx input]
+    (let [input (assoc input :del-statement-ids (query-statement-ids-by-actor tx input))]
```
cliffcaseyyet, line 138: i think this should be done one level up and then use the input framework.

src/db/sqlite/lrsql/sqlite/sql/delete.sql
```diff
@@ -59,3 +59,53 @@ DELETE FROM credential_to_scope
...
+
+-- :name delete-actor-st2actor!
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_actor where actor_ifi = :actor-ifi
+
+-- :name delete-actor-st2st!
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_statement WHERE descendant_id in (:del-statement-ids)
```
cliffcaseyyet, line 71: i do not think that IN clauses work with just a :del-statement-ids. We need to use a :v*: param to populate a list like that.

cliffcaseyyet:
> Remember that SQLite is only half of the SQL LRS backend; there's also the Postgres version that will need to be implemented.

let him get a working SQLite first, then go back. TBH this is almost all ANSI and one-liners so that port is going to be 30 minutes once all is said and done.

src/db/sqlite/lrsql/sqlite/sql/query.sql
cliffcaseyyet: Oh wait i see it a bit more clearly now. So this is to get the initial set then you'll have the delete fns weed out the dependencies.

src/main/lrsql/system/lrs.clj
```diff
@@ -199,6 +199,14 @@
...
         (actor-q/query-agent backend tx input))))
 
+  lrsp/AgentDelete
+  (-delete-person
+    [lrs _auth-identity params]
+    (let [conn (lrs-conn lrs)
+          input (agent-input/delete-actor-input params)]
+      (jdbc/with-transaction! [tx conn]
+        (stmt-cmd/delete-actor! backend tx input))))
+
```
milt, line 209: You'd probably want to declare this in the admin protocol https://github.com/yetanalytics/lrsql/blob/main/src/main/lrsql/admin/protocol.clj

src/db/sqlite/lrsql/sqlite/sql/delete.sql
milt: What Cliff said basically. We use soft-delete on reactions because statements have the FK, but in the other direction it's fine, the statement can just be deleted.

PR #330, opened 2023-08-28
SQL-201, actor-delete

invaliduser:
adds an agents/delete route that deletes an actor and associated reco…


src/db/postgres/lrsql/postgres/sql/delete.sql
```diff
@@ -61,3 +61,44 @@ DELETE FROM credential_to_scope
...
 AND secret_key = :secret-key
 AND scope = :scope::scope_enum;
+
+----------------------begin components of delete-actor-----
+-- :name delete-actor-st2st
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_statement 
+WHERE ancestor_id IN (query_goes_here)
+OR descendant_id IN (query_goes_here)
```
cliffcaseyyet, line 71: ?

src/main/lrsql/input/actor.clj
```diff
@@ -61,6 +61,17 @@
    :actor-ifi    actor-ifi
    :actor-type   actor-type})
 
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Actor Delete
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+#_(s/fdef delete-actor-input
```
cliffcaseyyet, line 69: ?

src/main/lrsql/admin/routes.clj
```diff
@@ -112,6 +112,14 @@
                              (ui/get-env inject-config))
      :route-name :lrsql.admin.ui/get-env]})
 
+(defn admin-delete-actor-route [common-interceptors jwt-secret jwt-leeway]
```
cliffcaseyyet, line 115: lets do something more like `admin-statement-operation-routes`?

src/main/lrsql/admin/protocol.clj
```diff
@@ -29,3 +29,6 @@
 (defprotocol AdminStatusProvider
   (-get-status [this params]
     "Get various LRS metrics."))
+
+(defprotocol ActorDeleter
```
cliffcaseyyet, line 33: Same name/group change as before I'd do this more somehting like `AdminStatementOperationProvider`

src/db/postgres/lrsql/postgres/sql/delete.sql
invaliduser: oops!  yeah, it's the same subquery as anywhere else

src/main/lrsql/admin/routes.clj
invaliduser: :heavy_check_mark:

src/main/lrsql/input/actor.clj
invaliduser: what *is* the play here? any munging necessary has been done, but I wanted to adhere to the input pattern.  Can remove this link in the chain if you think it better

cliffcaseyyet: Well what I was pointing out is it's commented out, we just want a valid spec here to wrap the input normalization. I would bet the spec is a little different for it to be valid, i.e. something like:
```
:args  (s/cat :actor-ifi :lrsql.spec.actor/actor-ifi])
  :ret   (s/keys :req-un [:lrsql.spec.actor/actor-ifi])
  ```

File: doc/endpoints.md [hunk 5]
```diff
@@ -33,5 +33,6 @@ The following examples use `http://example.org` as the URL body. All methods ret
 #### Misc Admin Routes
 
 - `GET http://example.org/admin/env`: Get select environment variables about the configuration which may aid in client-side operations. Currently returns a map containing the configuration variables `urlPrefix` and `enableStmtHtml`.
+- `DELETE http://example.org/agents`: Runs a *hard delete* of all records of an actor, and associated records (statements, attachments, etc).  Intended for privacy purposes like GDPR.
```
cliffcaseyyet, line 36: this route should have `/admin`

src/main/lrsql/backend/protocol.clj
```diff
@@ -38,6 +38,7 @@
   (-insert-actor! [this tx input])
   (-insert-statement-to-actor! [this tx input])
   (-update-actor! [this tx input])
+  (-wipe-actor! [this tx input])
```
cliffcaseyyet, line 41: let's just say delete lol

src/main/lrsql/admin/routes.clj
```diff
@@ -112,6 +112,14 @@
                              (ui/get-env inject-config))
      :route-name :lrsql.admin.ui/get-env]})
 
+(defn admin-statement-operation-routes [common-interceptors jwt-secret jwt-leeway]
+  #{["/agents" :delete (conj common-interceptors
```
cliffcaseyyet, line 116: see `/admin` prefix comment above.

src/main/lrsql/admin/protocol.clj
kelvinqian00: Chiming in: I actually prefer the general approach of Daniel's name since it's more concise. I do agree it should be more generic like `ActorManager` however.

kelvinqian00: I actually wonder if it can be combined with `AdminStatusProvider` since both deal with the LRS data that the admin has scope over.

cliffcaseyyet: i dunno. im envisioning a statement operations featureset (i know, its ugly but im sure there will be more), and status is literally a health check with some metrics.

kelvinqian00: So maybe `StatementManager` or `DataManager`? I just dislike the idea of naming the protocol an "operations provider" since it's very easy to confuse with the idea of providing data from the DB (i.e. what `AdminStatusProvider` does)

kelvinqian00: OTOH it could be useful to prefix the protocol name with "admin," as you suggested, since that would distinguish it from statement/actor ops that anyone can use (i.e. the xAPI spec).

cliffcaseyyet: That was my thinking exactly. It deliniates from normal statement ops and specifies admin statement ops. you can maybe combine the metrics one, thats fine, but i want the idea that its admin AND that its statement operations in there. if you dont like the word operations im open to alts.

kelvinqian00: Here's another option: `AdminResource`. Since the protocols defined by the lrs lib are `StatementsResource`, `DocumentResource`, and the like, so going with that scheme highlights that these are (admin) actions on the basic LRS data.

kelvinqian00: A couple more names:
- `AdminLRSManager`: Similar to the names of the other "admin manager" protocols.
- `AdminResource`: Goes with the precedent set by the lrs protocols like `StatementsResource` and `DocumentsResource`.

We'll let Daniel come up or select a name that he likes.

cliffcaseyyet: @kelvinqian00 `AdminLRSManager` nails it. Love it. @invaliduser you good w/that since its your feature?

src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -134,6 +134,15 @@
...
   (-update-actor! [_ tx input]
     (update-actor! tx input))
+  (-wipe-actor! [_ tx input]
+    (delete-actor-st2st input)
+    (delete-actor-st2activ input)
+    (delete-actor-attachments input)
+    (delete-actor-statements input)
+    (delete-actor-agent-profile input)
+    (delete-actor-state-document input)
+    (delete-actor-actor input))
```
kelvinqian00, line 144: I noticed here that `tx` is not being passed to the `delete-actor-*` functions. This is a serious error that will ensure that your delete ops will not work.

Despite not having added tests, the CI is already warning you about this. The linting CI job is emitting warnings about it:
```
src/db/postgres/lrsql/postgres/record.clj:103:20: warning: unused binding tx
src/db/sqlite/lrsql/sqlite/record.clj:137:20: warning: unused binding tx
```
Which you can see [here](https://github.com/yetanalytics/lrsql/actions/runs/6002993352/job/16280465970?pr=330). This is why you never turn off warnings: they may be actual errors!

BTW, if you want to know why we have a CI linter, Cliff and I use VSCode [Calva](https://calva.io/) which comes with the linter, [clj-kondo](https://github.com/clj-kondo/clj-kondo), preinstalled. We added this CI job so that people who use other editors can see the warnings we see.

kelvinqian00: (As a side note, the unit tests are failing, but for a completely different reason that I addressed in #331, so you can fix it by merging in the `main` branch.)

src/main/lrsql/admin/interceptors/lrs_management.clj
```diff
@@ -0,0 +1,26 @@
...
+                             (assoc ctx :actor-ifi params))))}))
+
+(def delete-actor
+  (interceptor
+   {:name ::delete-actor
+    :enter (fn delete-actor [ctx]
+             (let [{lrs :com.yetanalytics/lrs
+                    actor-ifi :actor-ifi}
+                   ctx]
+               (adp/-delete-actor lrs actor-ifi)))}))
```
milt, line 26: This interceptor is not returning the context/making a response

src/db/postgres/lrsql/postgres/sql/delete.sql
```diff
@@ -61,3 +61,51 @@ DELETE FROM credential_to_scope
 WHERE api_key = :api-key
 AND secret_key = :secret-key
 AND scope = :scope::scope_enum;
+
+----------------------begin components of delete-actor-----
+-- :name delete-actor-st2st
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_statement
```
cliffcaseyyet, line 69: FYI in Postgres i believe you can make this ALL the same hug fn, just use semicolons and put them one line after another. It should work to inject the `:actor-ifi` into all of them

milt: Would that have parity with the SQLite version though? Don't we generally match up the query fns?

cliffcaseyyet: no? I always saw it as we explicitly don't and that's the point of record.

src/db/postgres/lrsql/postgres/sql/delete.sql
invaliduser: fixed

src/main/lrsql/input/actor.clj
invaliduser: fixed (first arg is a one-key map, return value is the same)

src/main/lrsql/admin/protocol.clj
invaliduser: went with AdminLRSManager

doc/endpoints.md
invaliduser: fixed

src/main/lrsql/backend/protocol.clj
invaliduser: de-thesaurused

src/main/lrsql/admin/routes.clj
invaliduser: fixed

src/main/lrsql/admin/interceptors/lrs_management.clj
invaliduser: fixed

src/db/postgres/lrsql/postgres/sql/delete.sql
invaliduser: as these are clearly logically grouped w/the delete-actor prefix, I like Cliff's idea here

src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -358,3 +358,9 @@ ALTER TABLE activity ALTER COLUMN payload SET DATA TYPE JSONB;
 ALTER TABLE xapi_statement ALTER COLUMN payload SET DATA TYPE JSON;
 ALTER TABLE actor ALTER COLUMN payload SET DATA TYPE JSON;
 ALTER TABLE activity ALTER COLUMN payload SET DATA TYPE JSON;
+
+-- :name add-statement-to-actor-cascading-delete!
+-- :command :execute
+-- :doc Adds a cascading delete to delete st2actor entries when corresponding statements are deleted
+ALTER TABLE statement_to_actor DROP CONSTRAINT statement_fk;
+ALTER TABLE statement_to_actor ADD CONSTRAINT statement_fk FOREIGN KEY (statement_id) REFERENCES xapi_statement(statement_id) ON DELETE CASCADE;
```
cliffcaseyyet, line 366: I am not really seeing the idempotence in here. You need EITHER to check in another query if this is already done or do an if-not-exists kinda deal in this query. I see you dropping it but keep in mind it might take _minutes_ to populate an index on a sufficiently large db. This would mean that every startup takes minutes to run.

src/db/sqlite/lrsql/sqlite/record.clj
```diff
@@ -103,6 +103,8 @@
       (migrate-timestamps-activity-profile-03! tx)
       (migrate-timestamps-activity-profile-04! tx))
     (update-schema-simple! tx alter-credential-to-scope-scope-datatype!)
+    (when-not (= "CASCADE" (:on-delete (first (query-statement-to-actor-has-cascade-delete? tx))))
```
cliffcaseyyet, line 106: pertaining to above, it seems we do have the check here, so do similar for pg for sure.

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
```diff
@@ -426,3 +426,27 @@ ALTER TABLE activity_profile_document DROP COLUMN last_modified;
...
+  usage        TEXT CHECK (
+                 usage IN (''Actor'', ''Object'', ''Authority'', ''Instructor'', ''Team'',
+                           ''SubActor'', ''SubObject'', ''SubInstructor'', ''SubTeam'')
+               ) NOT NULL,                -- enum
+  actor_ifi    TEXT NOT NULL,             -- ifi string
+  actor_type   TEXT CHECK (
+                 actor_type IN (''Agent'', ''Group'')
+               ) NOT NULL,                -- enum
+
+    FOREIGN KEY (statement_id) REFERENCES xapi_statement(statement_id)
```
cliffcaseyyet, line 448: spacing looks weird.

src/db/sqlite/lrsql/sqlite/record.clj
invaliduser: duplicated (equivalent) to postgres

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
invaliduser: on second glance it actually seems nicely consistent to me, though I did rearrange it so `NOT NULL` always comes first

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,6 +183,49 @@
...
+
+        (assert (every? #(> % 0) (vals (get-stmt-#s))))
+        (doseq [ifi ifis]
+          (adp/-delete-actor lrs {:actor-ifi ifi}))
+        (is (every? #(= % 0) (vals (get-stmt-#s))))))
+    (let [stmt->ifi #(ua/actor->ifi (% "actor"))
+          count-of-actor (fn [actor-ifi] (-> (lrsp/-get-statements lrs auth-ident {:actor-ifi actor-ifi} []) :statement-result :statements count))
+          child-ifi (stmt->ifi (stmt-3 "object"))
+          parent-ifi (stmt->ifi stmt-3)]
+      (testing "substatement up"
```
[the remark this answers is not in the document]
invaliduser: clarified

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
cliffcaseyyet: What im asking is why is FK indented?

src/db/sqlite/lrsql/sqlite/record.clj
cliffcaseyyet: i am not seeing it in the latest commit

invaliduser: the logic is in record.clj rather than the SQL; I couldn't figure out how to do a conditional ALTER TABLE in plain Postgres SQL

https://github.com/yetanalytics/lrsql/blob/ca19a7e20d1ca68765b53d13565ed943271b811d/src/db/postgres/lrsql/postgres/record.clj#L70

src/test/lrsql/admin/protocol_test.clj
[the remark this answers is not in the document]
invaliduser: new language is:
 - "delete-actor correctly deletes statements that are parent to actor (sub)statements"
 - "delete-actor correctly deletes substatements that are child to actor statements"
 - "for StatementRefs, delete-actor deletes statement->actor relationships but leaves statements by another actor untouched"

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
invaliduser: unindented

src/db/postgres/lrsql/postgres/record.clj
```diff
@@ -66,7 +66,9 @@
       (migrate-agent-profile-document-last-modified! tx pd/local-tz-input))
     (if (-> tuning :config :enable-jsonb)
       (migrate-to-jsonb! tx)
-      (migrate-to-json! tx)))
+      (migrate-to-json! tx))
+    (when (= [] (check-statement-to-actor-cascading-delete tx))
```
cliffcaseyyet, line 70: this is bordering on nitpicky but in general is a good thing to know. If you annotate the SQL query with 
`-- :result :one`
it will return nil on no matches, which will make this a little cleaner than an empty vec comparison, i.e.
`(some? (check-statement-to-actor-cascading-delete tx))`

invaliduser: i am a nil supremacist, will change

src/db/postgres/lrsql/postgres/sql/ddl.sql
invaliduser: `add-statement-to-actor-cascading-delete!` is now gated on `check-statement-to-actor-cascading-delete` in `update-all!` of `postgres/record.clj`.  There will still be a re-indexing, but only once.

cliffcaseyyet (approved):

src/main/lrsql/admin/interceptors/lrs_management.clj
```diff
@@ -0,0 +1,30 @@
...
+            [io.pedestal.interceptor :refer [interceptor]]
+            [io.pedestal.interceptor.chain :as chain]
+            [lrsql.admin.protocol :as adp]))
+
+
+(def validate-delete-actor-params
+  (interceptor {:name ::validate-delete-actor-params
+                :enter (fn validate-delete-params [ctx]
+                         (let [params (get-in ctx [:request :json-params])]
+                           (if-some [err (s/explain-data (s/keys :req-un [:lrsql.spec.actor/actor-ifi]) params)]
```
milt, line 12: By convention we try to avoid inline spec defs like thisone. It might be good to take that `(s/keys ...)` and put it in a spec namespace, named as params for this operation (you can see other examples of this)

doc/endpoints.md
```diff
@@ -34,5 +34,6 @@ The following examples use `http://example.org` as the URL body. All methods ret
 #### Misc Admin Routes
 
 - `GET http://example.org/admin/env`: Get select environment variables about the configuration which may aid in client-side operations. Currently returns a map containing the configuration variables `urlPrefix` and `enableStmtHtml`.
+- `DELETE http://example.org/admin/agents`: Runs a *hard delete* of all records of an actor, and associated records (statements, attachments, etc).  Intended for privacy purposes like GDPR.
```
milt, line 37: Need some documentation of what the input param(s) are.

src/main/lrsql/input/actor.clj
```diff
@@ -61,6 +61,18 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Actor Delete
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(s/fdef delete-actor-input
+  :args (s/cat :actor-ifi (s/keys :req-un [:lrsql.spec.actor/actor-ifi]))
+  :ret  (s/keys :req-un [:lrsql.spec.actor/actor-ifi]))
+
+(defn delete-actor-input [actor-ifi-obj]
+  actor-ifi-obj)
```
milt, line 74: image

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -52,6 +53,33 @@
...
+   "actor"   {"mbox"       "mailto:sample.bar@example.com"
+              "objectType" "Agent"}
+   "verb"   {"id"      "http://adlnet.gov/expapi/verbs/asked"
+             "display" {"en-US" "asked"}}
+   "object"  {"objectType" "SubStatement"
+                       "actor"      {"mbox"       "mailto:sample.baz@example.com"
+                                     "objectType" "Agent"}
+                       "verb"       {"id"      "http://adlnet.gov/expapi/verbs/answered"
+                                     "display" {"en-US" "answered"}}
+                       "object"    {"id" "http://www.example.com/tincan/activities/multipart"}
```
milt, line 80: wacky formatting here

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,6 +183,49 @@
...
                 :result
                 (= :lrsql.admin/oidc-issuer-mismatch-error)))))
+    (testing "actor-delete"
+      (let [stmts [stmt-0 stmt-1 stmt-2 stmt-3]
+            ifis (->> (conj stmts (stmt-3 "object"))
+                      (map #(ua/actor->ifi (% "actor"))))
+            get-stmt-#s (fn []
+                          (reduce (fn [m actor-ifi]
+                                    (assoc m actor-ifi
+                                           (count (:statements (:statement-result (lrsp/-get-statements lrs auth-ident {:actor-ifi actor-ifi} []))))))
```
milt, line 193: try to keep to 80 chars when possible, avoid big lines like this

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,6 +183,49 @@
...
+                                  {} ifis))]
+        (lrsp/-store-statements lrs auth-ident stmts [])
+
+        (assert (every? #(> % 0) (vals (get-stmt-#s))))
```
milt, line 197: Why are these asserts rather than just `is`?

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,6 +183,49 @@
...
+      (testing "delete-actor correctly deletes substatements that are child to actor statements"
+        (lrsp/-store-statements lrs auth-ident [stmt-3] [])
+        (assert (= 1 (count-of-actor child-ifi)))
+        (adp/-delete-actor lrs {:actor-ifi parent-ifi})
+        (is (= 0 (count-of-actor child-ifi)))
+        (adp/-delete-actor lrs {:actor-ifi child-ifi}))
+      (testing "for StatementRefs, delete-actor deletes statement->actor relationships but leaves statements by another actor untouched"
+        (let [[ifi-0 ifi-2] (mapv stmt->ifi [stmt-0 stmt-2])]
+          (lrsp/-store-statements lrs auth-ident [stmt-0] [])
+          (lrsp/-store-statements lrs auth-ident [stmt-2] [])
```
milt, line 220: this could be a single call, might make this a little easier to read.

src/main/lrsql/input/actor.clj
invaliduser: i'm a people function!

src/test/lrsql/admin/protocol_test.clj
invaliduser: I thought it might be worth it to have a separation between "is the test setting up correctly" and "is this passing the tests"

src/test/lrsql/admin/protocol_test.clj
invaliduser: :+1:

src/test/lrsql/admin/protocol_test.clj
milt: That's an interesting idea, though I think letting the test output read everything out for you is probably better than wading through an assertion stack trace.

src/main/lrsql/input/actor.clj
milt: Yeah if it doesn't need an input function, I'd say remove it and just pass it along. It's been validated after all.

cliffcaseyyet: Well, so there is something to discuss here. Typically this is where the actual input objects are placed into an input map. Why is that not the case here? I see in record its expecting an input map, so is it pre-formed before getting here or something?

src/main/lrsql/system/lrs.clj
```diff
@@ -356,4 +356,11 @@
     (let [conn  (lrs-conn this)
           input (admin-stat-input/query-status-input params)]
       (jdbc/with-transaction [tx conn]
-        (admin-q/query-status backend tx input)))))
+        (admin-q/query-status backend tx input))))
+
+  adp/AdminLRSManager
+  (-delete-actor [this params]
+    (let [conn (lrs-conn this)
+          input (agent-input/delete-actor-input params)]
```
cliffcaseyyet, line 364: I see, you use the raw params map as a shortcut and presumably its got the same name. Hrmm, so technically this pulls anything from params and jams it into a db input. If we're trying to be consistent I would destructure params into `actor-ifi`, pass that into the input fn, and put it BACK on a map. Seems like extra work but it also does get rid of anything extraneous and ensure consistent format.

cliffcaseyyet: This will ensure proper worship of the cargo cult 😛

src/main/lrsql/admin/interceptors/lrs_management.clj
kelvinqian00: Seconded. We do not want any [magic specs](https://en.wikipedia.org/wiki/Magic_number_(programming)) in our code.

src/main/lrsql/input/actor.clj
invaliduser: the check takes place at ingestion in `admin.interceptors.lrs-management`; the spec for the route (`{"actor-ifi": <actor-ifi>}`) is identical with the spec for the op (`{:actor-ifi <actor-ifi>}`)

src/db/postgres/lrsql/postgres/sql/delete.sql
```diff
@@ -61,3 +61,30 @@ DELETE FROM credential_to_scope
...
+WHERE actor_ifi = :actor-ifi);
+
+DELETE FROM attachment WHERE statement_id IN (SELECT statement_id FROM statement_to_actor
+WHERE actor_ifi = :actor-ifi);
+
+DELETE FROM xapi_statement WHERE statement_id IN (SELECT statement_id FROM statement_to_actor
+WHERE actor_ifi = :actor-ifi);
+
+
+DELETE FROM agent_profile_document WHERE agent_ifi = :actor-ifi;
```
kelvinqian00, line 85: Extra space here?

src/db/postgres/lrsql/postgres/sql/delete.sql
```diff
@@ -61,3 +61,30 @@ DELETE FROM credential_to_scope
...
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_statement 
+WHERE ancestor_id IN (SELECT statement_id FROM statement_to_actor
+WHERE actor_ifi = :actor-ifi)
+OR descendant_id IN (SELECT statement_id FROM statement_to_actor
+WHERE actor_ifi = :actor-ifi);
+
+DELETE FROM statement_to_activity WHERE statement_id IN (SELECT statement_id FROM statement_to_actor
+WHERE actor_ifi = :actor-ifi);
```
kelvinqian00, line 76: So some suggestions on formatting:

I would start the subqueries on a separate line, to make it easier to tell where the main delete op ends and where the subquery begins. I would also put linebreaks before keywords like `WHERE` and `IN` in order to shorten line lengths and make the query easier (for humans) to parse. So I'd reformat the command like so:
```sql
DELETE FROM statement_to_activity
WHERE statement_id
IN (
  SELECT statement_id
  FROM statement_to_actor
  WHERE actor_ifi = :actor-ifi
);
```

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
```diff
@@ -426,3 +426,27 @@ ALTER TABLE activity_profile_document DROP COLUMN last_modified;
 -- :command :execute
 -- :doc Convert `activity_profile_document.last_modified` to timestamp - 04
 ALTER TABLE activity_profile_document RENAME COLUMN last_modified_tmp TO last_modified;
+
+--:name query-statement-to-actor-has-cascade-delete?
+select on_delete from pragma_foreign_key_list("statement_to_actor") where "table" ="xapi_statement"
```
kelvinqian00, line 431: Any reason why the keywords in this query are suddenly lowercase?

kelvinqian00: Also you missed a space after the `=` sign. Admittedly a very nitpicky comment.

src/main/lrsql/admin/interceptors/lrs_management.clj
```diff
@@ -0,0 +1,30 @@
+(ns lrsql.admin.interceptors.lrs-management
+  (:require [clojure.spec.alpha :as s]
+            [io.pedestal.interceptor :refer [interceptor]]
+            [io.pedestal.interceptor.chain :as chain]
+            [lrsql.admin.protocol :as adp]))
+
+
+(def validate-delete-actor-params
+  (interceptor {:name ::validate-delete-actor-params
```
kelvinqian00, line 9: It would be nice to start the `{:name ...` part on a new line to shorten line lengths and keep them under 80 columns.

Also did you forget to remove the extra spaces?

src/main/lrsql/input/actor.clj
cliffcaseyyet: Yeah fair enough. Seems fine.

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,6 +183,49 @@
...
+        (lrsp/-store-statements lrs auth-ident stmts [])
+
+        (assert (every? #(> % 0) (vals (get-stmt-#s))))
+        (doseq [ifi ifis]
+          (adp/-delete-actor lrs {:actor-ifi ifi}))
+        (is (every? #(= % 0) (vals (get-stmt-#s))))))
+    (let [stmt->ifi #(ua/actor->ifi (% "actor"))
+          count-of-actor (fn [actor-ifi] (-> (lrsp/-get-statements lrs auth-ident {:actor-ifi actor-ifi} []) :statement-result :statements count))
+          child-ifi (stmt->ifi (stmt-3 "object"))
+          parent-ifi (stmt->ifi stmt-3)]
```
kelvinqian00, line 204: I think you can put these `let` bindings in the outer `let`. That way you can get rid of one level of nesting.

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,6 +183,49 @@
...
+      (testing "delete-actor correctly deletes statements that are parent to actor (sub)statements"
```
kelvinqian00, line 205: So these tests are good progress, but I would like to see tests for everything that can be potentially deleted when you delete an actor. So that would mean:

- [x] Actor -> Statement
- [x] Actor -> Statement -> Parent Statement
- [x] Actor -> Statement -> Child Statement
- [ ] Actor -> Statement -> Activities
- [ ] Actor -> Statement -> Attachments
- [ ] Actor -> State Document
- [ ] Actor -> Agent Profile Document

src/test/lrsql/admin/protocol_test.clj
kelvinqian00: I agree with @milt. Using only `is` instead of sprinkling in `assert` makes the tests more readable, plus in my view it tells a story of "these statements/actors/etc used to be here, now they're not."

src/test/lrsql/admin/route_test.clj
```diff
@@ -291,6 +298,30 @@
...
+                               "objectType" "Agent"}
+                    "verb"    {"id"      "http://adlnet.gov/expapi/verbs/answered"
+                               "display" {"en-US" "answered"
+                                          "zh-CN" "回答了"}}
+                    "object"  {"id" "http://www.example.com/tincan/activities/multipart"}
+                    "context" {"platform" "example"}}
+            lrs (:lrs sys')
+            auth-ident {:agent  {"objectType" "Agent"
+                                 "account"    {"homePage" "http://example.org"
+                                               "name"     "12341234-0000-4000-1234-123412341234"}}
```
kelvinqian00, line 313: Same story with this agent (I think we may already have `def`s of agents, though I don't recall off the top of my head)

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,6 +183,49 @@
...
+                          (reduce (fn [m actor-ifi]
+                                    (assoc m actor-ifi
+                                           (count (:statements (:statement-result (lrsp/-get-statements lrs auth-ident {:actor-ifi actor-ifi} []))))))
+                                  {} ifis))]
+        (lrsp/-store-statements lrs auth-ident stmts [])
+
+        (assert (every? #(> % 0) (vals (get-stmt-#s))))
+        (doseq [ifi ifis]
+          (adp/-delete-actor lrs {:actor-ifi ifi}))
+        (is (every? #(= % 0) (vals (get-stmt-#s))))))
```
kelvinqian00, line 200: Fun fact: you can use `zero?` instead of `#(= % 0)`

src/main/lrsql/admin/interceptors/lrs_management.clj
invaliduser: spec now lives in `lrsql.spec.admin`

src/test/lrsql/admin/protocol_test.clj
invaliduser: re-jiggered

src/test/lrsql/admin/protocol_test.clj
invaliduser: switched to use of `is`

src/main/lrsql/system/lrs.clj
invaliduser: volcano god: sacrificed to

src/db/postgres/lrsql/postgres/sql/delete.sql
invaliduser: removed

src/db/postgres/lrsql/postgres/sql/delete.sql
invaliduser: amended as described

src/db/sqlite/lrsql/sqlite/sql/ddl.sql
invaliduser: capitalized, added

src/test/lrsql/admin/route_test.clj
```diff
@@ -291,6 +298,30 @@
...
         (is (empty? (select-keys headers sec-header-names)))))
+    (testing "delete actor route"
+      (let [stmt-0 {"id"      "00000000-0000-4000-8000-000000000000"
+                    "actor"   {"mbox"       "mailto:sample.foo@example.com"
+                               "objectType" "Agent"}
+                    "verb"    {"id"      "http://adlnet.gov/expapi/verbs/answered"
+                               "display" {"en-US" "answered"
+                                          "zh-CN" "回答了"}}
+                    "object"  {"id" "http://www.example.com/tincan/activities/multipart"}
+                    "context" {"platform" "example"}}
```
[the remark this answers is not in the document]
invaliduser: done

src/test/lrsql/admin/route_test.clj
invaliduser: done

src/test/lrsql/admin/protocol_test.clj
invaliduser: done

resources/lrsql/config/test/default/webserver.edn
```diff
@@ -19,6 +19,7 @@
                              ;; Allow access from admin-ui dev
                              "http://localhost:9500"]
  :url-prefix                "/xapi"
+ :enable-admin-delete-actor true
```
cliffcaseyyet, line 22: nice. true in dev but false in prod. I was gonna recommend this.

cliffcaseyyet (approved):

src/db/sqlite/lrsql/sqlite/sql/delete.sql
```diff
@@ -59,3 +59,46 @@ DELETE FROM credential_to_scope
...
+
+----------------------begin components of delete-actor-----
+-- :name delete-actor-st2st
+-- :command :execute
+-- :result :affected
+DELETE FROM statement_to_statement 
+WHERE ancestor_id IN (SELECT statement_id FROM statement_to_actor
+WHERE actor_ifi = :actor-ifi)
+OR descendant_id IN (SELECT statement_id FROM statement_to_actor
+WHERE actor_ifi = :actor-ifi)
```
kelvinqian00, line 71: I would like the re-formatting of the SQL delete ops to also apply to SQLite. I _thought_ this was obvious when I pointed it out in Postgres, but I guess not, so my apologies.

src/db/postgres/lrsql/postgres/sql/delete.sql
```diff
@@ -61,3 +61,39 @@ DELETE FROM credential_to_scope
...
+DELETE FROM xapi_statement WHERE statement_id IN (
+  SELECT statement_id FROM statement_to_actor
+  WHERE actor_ifi = :actor-ifi
+);
+
+DELETE FROM agent_profile_document WHERE agent_ifi = :actor-ifi;
+
+DELETE FROM state_document WHERE agent_ifi = :actor-ifi;
+
+DELETE FROM actor where actor_ifi = :actor-ifi;
```
kelvinqian00, line 98: Apologies for not noticing the first time, but this `where` should also be capitalized by convention.

src/db/sqlite/lrsql/sqlite/sql/delete.sql
```diff
@@ -59,3 +59,46 @@ DELETE FROM credential_to_scope
...
+
+-- :name delete-actor-state-document
+-- :command :execute
+-- :result :affected
+DELETE FROM state_document WHERE agent_ifi = :actor-ifi
+
+-- :name delete-actor-actor
+-- :command :execute
+-- :result :affected
+DELETE FROM actor where actor_ifi = :actor-ifi
```
kelvinqian00, line 104: Same here

src/test/lrsql/admin/protocol_test.clj
kelvinqian00: I don't see the change?

kelvinqian00: Also the `#(> % 0)` that's immediately above can be replaced with `pos-int?`

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,7 +187,79 @@
...
+          (is (zero? (count-of-actor child-ifi)))
+          (adp/-delete-actor lrs {:actor-ifi child-ifi}))
+        (testing "for StatementRefs, delete-actor deletes statement->actor relationships but leaves statements by another actor untouched"
+          (let [[ifi-0 ifi-2] (mapv stmt->ifi [stmt-0 stmt-2])]
+            (lrsp/-store-statements lrs auth-ident [stmt-0 stmt-2] [])
+            (is (= 2 (count-of-actor ifi-0)))
+            (is (= 2 (count-of-actor ifi-2)))
+            (adp/-delete-actor lrs {:actor-ifi ifi-0})
+            (is (= 1 (count-of-actor ifi-0)))
+            (is (= 1 (count-of-actor ifi-2)))
```
kelvinqian00, line 232: Actually @cliffcaseyyet can you confirm that leaving deleted IFIs in statements is okay as long as they're referenced by a Statement Ref, right?

cliffcaseyyet: @kelvinqian00 That's not exactly what is happening afaik. What we decided was if A (with Actor X) references B (with Actor Y) and we delete Actor Y, statement B gets deleted but Statement A remains in place with a dead ref (apparently legal in xAPI).

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,7 +187,79 @@
...
+      (lrsp/-set-document lrs auth-ident lrst/state-id-params lrst/state-doc-1 true)
+      (adp/-delete-actor lrs {:actor-ifi ifi})
+      (is (empty? (jdbc/execute! ds ["select * from state_document where agent_ifi  = ?" ifi])))))
+
+    (testing "delete-actor handles dependent tables: agent_profile_document"
+      (let [ifi (ua/actor->ifi (:agent lrst/agent-prof-id-params))
+            ds (-> lrs :connection :conn-pool)]
+        (lrsp/-set-document lrs auth-ident lrst/agent-prof-id-params lrst/agent-prof-doc true)
+        (adp/-delete-actor lrs {:actor-ifi ifi})
+        (is (empty? (jdbc/execute! ds ["select * from agent_profile_document where agent_ifi  = ?" ifi])))))
```
kelvinqian00, line 249: You can use the [get-document](https://github.com/yetanalytics/lrsql/blob/SQL-201/src/main/lrsql/system/lrs.clj#L169) protocol function instead of hand-writing a query. You would have to specify a `:stateId` to query the state document, and you would have to specify a `:profileId` and `:agent` to query the agent profile document.

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,7 +187,79 @@
...
+
+    (let [ifi (ua/actor->ifi (lrst/stmt-4 "actor"))
+          stmt-id (u/str->uuid (lrst/stmt-4 "id"))
+          ds (-> lrs :connection :conn-pool)]
+      (lrsp/-store-statements lrs auth-ident [lrst/stmt-4] [lrst/stmt-4-attach])
+      (adp/-delete-actor lrs {:actor-ifi ifi})
+
+      (testing "delete-actor handles statement attachments"
+        (is (empty? (jdbc/execute! ds ["select * from attachment where statement_id  = ?" stmt-id]))))
```
kelvinqian00, line 258: If you query a statement, [you will get an `:attachments` value as part of the result value](https://github.com/yetanalytics/lrsql/blob/SQL-201/src/main/lrsql/ops/query/statement.clj#L89). And if you want to go lower-level without having to manually write out a statement, there is the `backend/-query-attachments` query (though you would need to grab the current DB backend record from the system).

src/test/lrsql/admin/protocol_test.clj
kelvinqian00: I don't see a test for the actor itself? (Which I find odd since wouldn't that be the _simplest_ thing to test?)

src/test/lrsql/admin/protocol_test.clj
```diff
@@ -155,7 +187,79 @@
...
+      (testing "delete actor handles intermediate statement-to-activity table"
+        (is (empty? (jdbc/execute! ds ["select * from statement_to_activity where statement_id  = ?" stmt-id])))))
```
kelvinqian00, line 260: Okay this I'll slide since there's no query function to query the join table itself. I would extract this part into a helper function though.

src/main/lrsql/input/actor.clj
```diff
@@ -61,6 +61,18 @@
    :actor-ifi    actor-ifi
    :actor-type   actor-type})
 
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Actor Delete
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(s/fdef delete-actor-input
```
kelvinqian00, line 69: Weird spacing with the blank lines on lines 64 and 65, and the lack thereof after line 68?

src/test/lrsql/admin/protocol_test.clj
kelvinqian00: Doesn't seem like you made that change

src/test/lrsql/admin/protocol_test.clj
invaliduser: There is one now.

src/test/lrsql/admin/protocol_test.clj
invaliduser: Ah, yes, I'd made the change elsewhere but forgotten the original case.  Changed.

src/db/sqlite/lrsql/sqlite/sql/delete.sql
invaliduser: changed

src/db/postgres/lrsql/postgres/sql/delete.sql
invaliduser: capitalized

src/db/sqlite/lrsql/sqlite/sql/delete.sql
invaliduser: capitalized

src/test/lrsql/admin/protocol_test.clj
invaliduser: switched to use of `get-document`

src/test/lrsql/admin/protocol_test.clj
invaliduser: switched to checking `attachments` key

src/main/lrsql/input/actor.clj
invaliduser: adjusted

PR #335, opened 2023-09-01
sleep long cast fix for jdk16+

cliffcaseyyet:
sleep long cast fix for jdk16+

### DIFF
--- a/src/main/lrsql/util/concurrency.clj
+++ b/src/main/lrsql/util/concurrency.clj
@@ -50,7 +50,7 @@
       (if (and (< attempt max-attempt)
                (retry-test e))
         (let [sleep (backoff-ms (inc attempt) opts)]
-          (Thread/sleep sleep)
+          (Thread/sleep (long sleep))
           (rerunable-txn* txn-expr (inc attempt) opts))
         (do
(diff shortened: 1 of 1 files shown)


src/main/lrsql/util/concurrency.clj
```diff
@@ -50,7 +50,7 @@
       (if (and (< attempt max-attempt)
                (retry-test e))
         (let [sleep (backoff-ms (inc attempt) opts)]
-          (Thread/sleep sleep)
+          (Thread/sleep (long sleep))
```
kelvinqian00, line 53: I'd add a comment because this is super not obvious to somebody who is not familiar with the inner workings of Clojure/Java (i.e. `Thread/sleep` in JDK 16+ does not accept Clojure's boxed primitives)

As a super side note, out of curiosity do primitive type hints also work?

kelvinqian00: I noticed in `backoff-ms` that you coerce to `int` in the `:else` case, plus you return `nil` in the `(> attempt max-attempt)` case.

milt (approved):

kelvinqian00 (approved):

PR #338, opened 2023-09-18
Suppress CVE-2023-40167

kelvinqian00:
This CVE is actually listed has having "no known exploit scenario" (plus LRS validation should prevent issues anyways)


milt (approved):
Looks good. Don't forget to also address the corresponding Dependabot alert: https://github.com/yetanalytics/lrsql/security/dependabot

PR #348, opened 2023-11-20
Update Pedestal to v0.6.2 to address CVEs

kelvinqian00:
Update Pedestal to address [these CVEs](https://github.com/pedestal/pedestal/issues/763).

NOTE: For now this keeps Jetty at v9.x.x, but in the long term Pedestal's Jetty version will be updated to v11.x.x, so we will need to prepare for this eventuality.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -24,22 +24,7 @@
   com.zaxxer/HikariCP                      {:mvn/version "5.0.0"
                                             :exclusions  [org.slf4j/slf4j-api]}
   ;; Pedestal and Jetty webserver deps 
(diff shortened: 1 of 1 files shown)


milt:
> Update Pedestal to address [these CVEs](https://github.com/pedestal/pedestal/issues/763).
> 
> NOTE: For now this keeps Jetty at v9.x.x, but in the long term Pedestal's Jetty version will be updated to v11.x.x, so we will need to prepare for this eventuality.

Hey that's good news

milt (approved):

cliffcaseyyet (approved):

PR #366, opened 2024-01-26
[SQL-170] Implement profile scopes

kelvinqian00:
Implement `activities_profile` and `agents_profile` credential scopes, similar to the `profile` scope recommended in [this section](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Communication.md#details-15) of the xAPI spec. We implement those two scopes instead of the `profile` scope directly since:
- there is already a reserved `profile` keyword for OIDC
- it would allow the user to be more granular with their document scopes

In addition, we add a `lrsql.scope-test` namespace for HTTP tests on different credential scopes (as different scopes are not covered by the xAPI conformance tests).


src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -281,6 +280,35 @@ CREATE TYPE scope_enum AS ENUM (
...
 ALTER TABLE IF EXISTS credential_to_scope ALTER COLUMN scope TYPE scope_enum USING (scope::scope_enum);
 
+/* Migration 2024-01-24 - Add profile and profile/read scopes */
+
+/* Simply adding the profile scope will cause a name clash with the reserved
+   OIDC profile scope. As a result, we add prefixes to create activity_profile
+   and agent_profile (and their read-only versions); this has the additional
+   benefit of narrowing the scope to just activity and agent profiles,
+   respectively. Since profile and profile/read have always been unused, we
+   are safe to remove them as enums. */
```
kelvinqian00, line 290: Would it be better to simply change the initial migration function (and note the change in the comments) instead of adding an additional migration function (and thus overhead)?

src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -281,6 +280,35 @@ CREATE TYPE scope_enum AS ENUM (
...
+  'all/read',
+  'all',
+  'state',
+  'state/read',
+  'define',
+  -- new; add document/profile scopes (also document/state out of being pedantic)
+  'activities_profile',
+  'activities_profile/read',
+  'agents_profile',
+  'agents_profile/read');
```
kelvinqian00, line 310: Not sure which is better: the plural `activities_profile`/`agents_profile` or the singular `activity_profile`/`agent_profile`

src/db/postgres/lrsql/postgres/sql/ddl.sql
milt: Generally it is good practice to keep migrations isolated in to their own code. This becomes a factor if something gets mucked up in production and a database needs to be fixed from the REPL.

src/db/postgres/lrsql/postgres/sql/ddl.sql
milt: I would go with singular

src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -281,6 +280,35 @@ CREATE TYPE scope_enum AS ENUM (
...
+   and agent_profile (and their read-only versions); this has the additional
+   benefit of narrowing the scope to just activity and agent profiles,
+   respectively. Since profile and profile/read have always been unused, we
+   are safe to remove them as enums. */
+
+-- :name alter-scope-enum-type-2!
+-- :command :execute
+-- :doc Add `activity_profile`, `activity_profile/read`, `agent_profile`, and `agent_profile/read` to `credential-to-scope.scope` enum.
+ALTER TABLE IF EXISTS credential_to_scope ALTER COLUMN scope TYPE VARCHAR(255);
+DROP TYPE IF EXISTS scope_enum;
```
milt, line 296: I assume this preserves existing references to the scopes? I'm pretty sure they just get stored as ints on the rows, but just wanted to be sure.

src/db/postgres/lrsql/postgres/record.clj
```diff
@@ -54,6 +54,7 @@
     (alter-admin-account-passhash-optional! tx)
     (alter-admin-account-add-openid-issuer! tx)
     (alter-scope-enum-type! tx)
+    (alter-scope-enum-type-2! tx)
```
milt, line 57: As I said below, this should be it's own migration function, probably with a more descriptive name. I'd also wrap it with a query to see if it needs to be executed.

src/db/postgres/lrsql/postgres/sql/ddl.sql
kelvinqian00: I forgot the exact reason this was added in the first enum migration, but that's probably the reason why. I think if you don't include that line it would be similar to deleting table that is referenced by foreign keys - it'll get blocked.

src/db/postgres/lrsql/postgres/record.clj
kelvinqian00: Not sure what would be a more descriptive name. I guess the query wrapping would reduce overhead but I'm not sure if the overhead saved would be more than the extra overhead needed for the query.

Now that I'm looking at the function pair, it might be a good idea to _replace_ the original `alter-scope-enum-type!` with `alter-scope-enum-type-2!`, since the former's changes will get overwritten anyways.

src/db/postgres/lrsql/postgres/sql/ddl.sql
kelvinqian00: Actually I realized that these scope names follow from the names of their respective resource paths. So the `statements` scope matches the `/statements` resource, and the `state` scope matches the `/state` scope. By that logic, the `activities_profile` and `agents_profile` scopes should follow the `/activities/profile` and `/agents/profile` resources (and the hypothetical scopes `activities` and `agents` the resources `/activities` and `/agents`).

milt: yeah that justification is sound to me

src/db/postgres/lrsql/postgres/sql/ddl.sql
milt: I just want to be certain that, whatever we do, we don't break existing data if an LRS updates.

kelvinqian00: Given that these lines are present in the first migration, I have my doubts that stuff would break.

src/db/postgres/lrsql/postgres/record.clj
kelvinqian00: Did the replacement in ce69659e28bc5efad3c222d345ce50bda9e7a957 after performing the integration tests needed to see that the migration preserves existing scopes. Also renamed each function to `*-v2!` in 6287914af37b81f16d06c42bee2166fc182c88da to highlight the fact that the original migration functions are now superseded.

milt (approved):

PR #369, opened 2024-02-02
SQL-231 catch token exceptions in no-val mode

milt:
[SQL-231] If no-val mode is used and no token is present or is malformed we can get an NPE and a 500 response. This change puts a try/catch around the token decoding code that leads to a 401 if there is not a valid token, rather than the 500.

[SQL-231]: https://yet.atlassian.net/browse/SQL-231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/lrsql/admin/interceptors/jwt.clj
+++ b/src/main/lrsql/admin/interceptors/jwt.clj
@@ -31,7 +31,9 @@
                          (if (some? username)
                            (:result (adp/-ensure-account-oidc lrs username issuer))
                            result))
-                       (catch Exception _
+                       (catch Exception ex
+                         ;; We want any error here to return a 401, but we log
+                         (log/warnf ex "No-val JWT Error: %s" (ex-message ex))
                          :lrsql.admin/unauthorized-token-error))
                      ;; normal jwt, check signature etc
                      (admin-u/jwt->account-id token secret leeway))]
--- a/src/main/lrsql/admin/interceptors/jwt.clj
+++ b/src/main/lrsql/admin/interceptors/jwt.clj
@@ -31,7 +31,7 @@
                          (if (some? username)
                            (:result (adp/-ensure-account-oidc lrs username issuer))
                            result))
-                       (catch Exception ex_
+                       (catch Exception _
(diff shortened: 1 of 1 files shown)


src/main/lrsql/admin/interceptors/jwt.clj
```diff
@@ -23,13 +23,16 @@
...
-                         result))
+                     (try
+                       (let [{:keys [issuer username] :as result}
+                             (admin-u/proxy-jwt->username-and-issuer
+                              token no-val-uname no-val-issuer no-val-role-key
+                              no-val-role)]
+                         (if (some? username)
+                           (:result (adp/-ensure-account-oidc lrs username issuer))
+                           result))
+                       (catch Exception ex_
```
kelvinqian00, line 34: Doesn't the underscore usually come _before_ the ident, i.e. `_ex`?

kelvinqian00: Looks the linter caught it first.

src/main/lrsql/admin/interceptors/jwt.clj
```diff
@@ -23,13 +23,16 @@
...
+                     (try
+                       (let [{:keys [issuer username] :as result}
+                             (admin-u/proxy-jwt->username-and-issuer
+                              token no-val-uname no-val-issuer no-val-role-key
+                              no-val-role)]
+                         (if (some? username)
+                           (:result (adp/-ensure-account-oidc lrs username issuer))
+                           result))
+                       (catch Exception _
+                         :lrsql.admin/unauthorized-token-error))
```
kelvinqian00, line 35: I think the Exception can be made more specific. Otherwise any old exception can bubble up and be interpreted as a 401, e.g. if there's something wrong with the SQL layer.

kelvinqian00: Discussed in Slack - this is a special case to catch all exceptions as 401 since this is auth

kelvinqian00 (approved):

PR #372, opened 2024-02-09
adds openapi spec to resources/lrsql/doc and a route to serve it in json form

invaliduser:


resources/lrsql/doc/openapi.json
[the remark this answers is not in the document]
milt: And ensuring a newline

src/main/lrsql/admin/interceptors/openapi.clj
```diff
@@ -0,0 +1,11 @@
...
+  (:require [io.pedestal.interceptor :refer [interceptor]]))
+
+(def openapi
+  "Returns an openapi spec in json form for lrsql"
+  (interceptor
+   {:name ::openapi
+    :enter
+    (fn openapi [_ctx]
+      {:status 200
+       :body (slurp "/resources/doc/openapi.json")})}))
```
milt, line 11: It looks like this will hit the filesystem on every request. That might be desirable in development when working on the schema itself but we probably want to def it into memory and just serve that.

src/main/lrsql/admin/routes.clj
```diff
@@ -169,13 +170,21 @@
...
 (defn admin-lrs-management-routes [common-interceptors jwt-secret jwt-leeway no-val-opts]
   #{["/admin/agents" :delete (conj common-interceptors
                                    lm/validate-delete-actor-params
                                    (ji/validate-jwt jwt-secret jwt-leeway no-val-opts)
                                    ji/validate-jwt-account
                                    lm/delete-actor)
-     :route-name :lrsql.lrs-management/delete-actor]})
+     :route-name :lrsql.lrs-management/delete-actor]
+    ["/admin/openapi" :get  (conj common-interceptors
+
```
milt, line 184: there's a blank line hanging out here

src/main/lrsql/admin/routes.clj
```diff
@@ -169,13 +170,21 @@
...
 (defn admin-lrs-management-routes [common-interceptors jwt-secret jwt-leeway no-val-opts]
   #{["/admin/agents" :delete (conj common-interceptors
                                    lm/validate-delete-actor-params
                                    (ji/validate-jwt jwt-secret jwt-leeway no-val-opts)
                                    ji/validate-jwt-account
                                    lm/delete-actor)
-     :route-name :lrsql.lrs-management/delete-actor]})
+     :route-name :lrsql.lrs-management/delete-actor]
+    ["/admin/openapi" :get  (conj common-interceptors
```
milt, line 183: the path should be versioned, what you've got there should be v0

src/main/lrsql/admin/routes.clj
```diff
@@ -169,13 +170,21 @@
...
+                                   
+                                   (ji/validate-jwt jwt-secret jwt-leeway no-val-opts)
+                                   ji/validate-jwt-account
```
milt, line 186: Do we want the openapi endpoint to be authenticated?

src/main/lrsql/admin/routes.clj
invaliduser: :+1:

resources/lrsql/doc/openapi.yaml
```diff
@@ -0,0 +1,790 @@
...
+          type: array
+          items: {type: string}
+        openid*:
+          type: array
+          items: {$ref: '#/components/schemas/URI'}
+        account*:
+          type: array
+          items: {$ref: '#/components/schemas/Account'}
+      required: [objectType]
+    Statement: {type: object, description: 'https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#20-statements'}
```
milt, line 121: this might be a good place for a `$comment` about the fact that the statement spec is intentionally omitted

src/main/lrsql/admin/interceptors/openapi.clj
invaliduser: I debated about this---do we?  I had a vision of lrs sitting there in production, all nicely configured, but w/the file in memory *forever*.  Happy to memoize it, but...do we?

src/main/lrsql/admin/routes.clj
invaliduser: I thought about this---since they need auth to make any actual requests, didn't seem like it was actually denying anything useful.  Happy to open it up if we want.

src/main/lrsql/admin/interceptors/openapi.clj
cliffcaseyyet: @invaliduser Yes because it's looking in resources classpath for the file. We are not gonna do surgery on running Jars to fix doc. Additionally, changing the docs means changing the API, and that by definition cannot be done without a restart. On the flip side i don't care passionately about the performance of a non-operational endpoint. Overall I vote Memo.

PR #383, opened 2024-03-21
SQL-237 - Enable Custom Text Overrides AND custom localization support

cliffcaseyyet:
Serve language override file and lang pref as part of /env


resources/lrsql/customization/language.json
kelvinqian00, line 1: This file will need some documentation on how to use.

Makefile
```diff
@@ -169,6 +169,10 @@ target/bundle/config/authority.json.template.example:
 
 target/bundle/config: target/bundle/config/lrsql.json.example target/bundle/config/authority.json.template.example
 
+target/bundle/customization:
```
milt, line 172: in order to work this would need to be _at least_ `target/bundle/lrsql/customization` to maintain the relative resource path. As I mentioned on slack I'd also favor an additional wrapping directory so just that can be included in classpath overrides.

src/main/lrsql/init/customization.clj
```diff
@@ -0,0 +1,15 @@
+(ns lrsql.init.customization
```
milt, line 1: I would suggest making this namespace (and the corresponding resource dir) "localization" or something like that, to differentiate between general customization in the release

milt: In that scheme the file in the release would be like `/customization/lrsql/localization/language.json` and the classpath arg would be `-cp customization`

milt: Something like the FE index doc would be `/customization/public/admin/index.html` could be included with the same classpath override

doc/env_vars.md
```diff
@@ -201,5 +201,6 @@ These config vars enable and configure browser security response headers from th
...
 | `LRSQL_ENABLE_ADMIN_STATUS`       | `enableAdminStatus`      | Whether or not to serve admin system status data that queries the database.                                                                                                                                                                            | `true`                                                                                     |
 | `LRSQL_ENABLE_STMT_HTML`          | `enableStmtHtml`         | Whether or not HTML data is returned in the LRS HTTP response. If `false` disables HTML rendering even if `LRSQL_ENABLE_ADMIN_UI` is `true`. In that case the UI will not display the Statement Browser feature.                                       | `true`                                                                                     |
+| `LRSQL_ADMIN_LANGUAGE`          | `adminLanguage`         | Allows override of default RFC5646 language code (`en-US`). May require that you add new language map entries to `resources/customization/language.json` to add a new language.                                       | `en-US`                                                                                     |
```
milt, line 204: you're missing the `lrsql` segment here too

resources/lrsql/customization/language.json
kelvinqian00: Discussed on Slack why this course of action will not be taken.

milt (approved):

kelvinqian00 (approved):

src/main/lrsql/init/localization.clj
```diff
@@ -0,0 +1,14 @@
+(ns lrsql.init.localization
+  "Utilities for white labeling/localization overrides"
+  (:require [cheshire.core :as json]
+            [clojure.java.io :as io]))
+
+(def lang-path
+  "lrsql/localization/language.json")
+
+(def custom-language-map
```
milt, line 9: As discussed on slack, put this in a memoized function to call when serving the `GET /env` request. That should make this safe for AOT and still allow override at runtime.

PR #396, opened 2024-04-24
SQL-244 Support for table view in data browser

cliffcaseyyet:
- Added statement get max to env call

### DIFF
--- a/src/main/lrsql/system/webserver.clj
+++ b/src/main/lrsql/system/webserver.clj
@@ -64,7 +64,7 @@
          config
          (:config lrs))
         ;; LRS reaction toggle
-        {:keys [enable-reactions]}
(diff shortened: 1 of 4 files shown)


milt (approved):

src/main/lrsql/system/webserver.clj
```diff
@@ -64,7 +64,7 @@
          config
          (:config lrs))
         ;; LRS reaction toggle
-        {:keys [enable-reactions]}
```
kelvinqian00, line 67: This is no longer _just_ about LRS reactions toggling, isn't it ;)

cliffcaseyyet: fixed, but i would really focus in the UI one ;)

kelvinqian00 (approved):

PR #397, opened 2024-04-24
Cft updates

invaliduser:


docker-compose.yml
```diff
@@ -10,9 +10,7 @@ services:
       POSTGRES_USER: lrsql_user
```
cliffcaseyyet, line 10: I am not sure that we want to make ALL uses of docker-compose be clustered. I would consider dumping this one into dev-resources.

dev-resources/template/2_lrs.yml
```diff
@@ -77,7 +77,11 @@ Parameters:
     Description: Version of SQL LRS to download and install (public release versions on GitHub)
     Type: String
     Default: v0.5.5
-
+  JWTCommonSecret:
```
cliffcaseyyet, line 80: Need path and version

dev-resources/template/2_lrs.yml
```diff
@@ -515,7 +519,8 @@ Resources:
                     "webserver": {
                       "httpHost": "0.0.0.0",
                       "httpPort": ${InstanceHttpPort},
-                      "allowedOrigins": ${AllowedOrigins}
+                      "allowedOrigins": ${AllowedOrigins},
+                      "jwtCommonSecret": ${JWTCommonSecret}
```
cliffcaseyyet, line 523: This is for sure not gonna work. This would just take the literal path that get input and print it, take a look here:

https://github.com/yetanalytics/lrsql/blob/main/dev-resources/template/1_db.yml#L127

Unfortunately it's possible you will need to get a lambda to decrypt it, or optionally we can just forget about SSM storage because it's going to be a complete PITA to decrypt. Try doing what was done in the other template first though.

cliffcaseyyet: Yet another option is to NOT store it as a secure string, but to store it in SSM as a parameter. In that case it will be writable to this file via reference. I'm pretty sure SSM with secure strings needs a lambda decrypter which is complicated for any of us to do. @milt thoughts?

cliffcaseyyet: see slack.

PR #402, opened 2024-05-08
Generates openapi spec from annotated routes

invaliduser:
- annotates routes with [gen-openapi](https://github.com/yetanalytics/gen-openapi/)
 - combines with annotations from similarly-annotated lrs routes
 - serves generated openAPI spec in JSON form at /admin/openapi


src/main/lrsql/system/webserver.clj
```diff
@@ -162,6 +163,8 @@
                           i/xapi-default-interceptors
                           http/create-server
                           http/start)]
+          (reset! openapi/oa-routes (::http/routes service))
```
milt, line 166: What's the rationale for using state like this vs making a function to build the interceptor?

invaliduser: yeah, for simplicity's sake that makes sense, will change

src/main/lrsql/system/openapi.clj
```diff
@@ -0,0 +1,12 @@
+(ns lrsql.system.openapi
```
milt, line 1: Note the other `system` namespaces are all for `component`s. It's a little counter-intuitive, but in the convention of the project stuff like this would go under an `init` ns

src/main/lrsql/system/webserver.clj
invaliduser: actually I take this back, it ends up being a circular dependency to pass it through `add-admin-routes`

milt: OK, was wondering about that. I'm gonna take a closer look at the code

src/main/lrsql/admin/interceptors/lrs_management.clj
milt, line 1: Is lrs_management a good place for this? I think it could have its own ns

kelvinqian00:
It would be nice to have the endpoints alphabetically sorted. This is especially important to visually separate the `/admin` endpoints and the `/xapi` endpoints when eyeballing them on Swagger.

src/main/lrsql/system/openapi.clj
```diff
@@ -0,0 +1,12 @@
...
+  (:require 
+   [com.yetanalytics.gen-openapi.generate.schema :as gs]
+   [com.yetanalytics.lrs.pedestal.openapi :as lrsoa]))
+
+(def oa-routes (atom nil))
+(def general-map {:openapi "3.0.0"
+                  :info {:title "LRSQL"
+                         :version "0.7.2"}
```
cliffcaseyyet, line 9: is this supposed to be the release version? It's kinda rough when the release version needs to be manually updated on each push. There ARE solutions where we can introspect git repo on build for this, but they can be hairy.

src/main/lrsql/admin/interceptors/lrs_management.clj
invaliduser: moved

src/main/lrsql/system/openapi.clj
invaliduser: moved

src/main/lrsql/system/webserver.clj
invaliduser: ended up adding another step to construction of `routes`

src/main/lrsql/admin/interceptors/openapi.clj
```diff
@@ -0,0 +1,20 @@
...
+             (assoc ctx :response
+                    {:status 200
+                     :body (oa-core/make-oa-map
+                            {:openapi "3.0.0"
+                             :info {:title "LRSQL"
+                                    :version version}
+                             :externalDocs {:url "https://github.com/yetanalytics/lrsql/blob/main/doc/endpoints.md"}
+                             :components (gs/dsl (add-lrsql-specifics
+                                                  lrs-oa/components))}
+                            routes)}))}))
```
milt, line 20: Since you already have this wrapped inside a function, why not do this in a let binding above so it just happens the once on startup?

src/main/lrsql/system/webserver.clj
```diff
@@ -109,6 +110,18 @@
...
+        routes (add-openapi-route
+                routes
+                {:lrs lrs
+                 :head-opts {:sec-head-hsts         sec-head-hsts
+                             :sec-head-frame        sec-head-frame
+                             :sec-head-content-type sec-head-content-type
+                             :sec-head-xss          sec-head-xss
+                             :sec-head-download     sec-head-download
+                             :sec-head-cross-domain sec-head-cross-domain
+                             :sec-head-content      sec-head-content}
```
milt, line 122: maybe let-bind the head-opts map above so you can just refer to it here and at the other site above

Makefile
```diff
@@ -207,6 +216,12 @@ target/bundle/admin: resources/public/admin
 	mkdir -p target/bundle
 	cp -r resources/public/admin target/bundle/admin
 
+# Include version file
+target/bundle/version:
+	mkdir -p target/bundle
+	make git-details.json
+	mv git-details.json target/bundle/git-details.json
+
```
milt, line 224: I don't get why it isn't just a single make target like `target/bundle/git-details.json`

src/main/lrsql/admin/interceptors/openapi.clj
invaliduser: :+1:

src/main/lrsql/system/openapi.clj
milt: where to?

src/main/lrsql/admin/interceptors/lrs_management.clj
milt: where to?

Makefile
invaliduser: I wanted to keep it convenient for dev  to just move into `dev/resources`

src/main/lrsql/init/git_data.clj
```diff
@@ -0,0 +1,30 @@
...
+  (:require [cheshire.core :as json]
+            [clojure.string]))
+
+(defn read-git []
+  (let [read-text (or (try (slurp "target/bundle/git-details.json")
+                           (catch Exception _
+                             nil))
+                      (try (slurp "dev-resources/git-details.json")
+                           (catch Exception _
+                             nil)))]
```
milt, line 11: The production path here won't work and the dev one won't unless you do a manual copy, I think we need to work on this bit.

src/main/lrsql/admin/interceptors/lrs_management.clj
invaliduser: `src/main/lrsql/admin/interceptors/openapi.clj`

src/main/lrsql/system/webserver.clj
invaliduser: done

src/main/lrsql/init/git_data.clj
invaliduser: what path would work here for prod?  This is a jvm thing I'm afraid I don't understand.

milt: Let's go over it on slack

src/main/lrsql/init/git_data.clj
```diff
@@ -0,0 +1,10 @@
+(ns lrsql.init.git-data
+  (:require [clojure.edn :as edn]
+            [clojure.string]
+            [clojure.java.shell :refer [sh]]))
+
+(defn read-version []
+  (try (edn/read-string (slurp "lrsql/config/git-details.edn"))
+       (catch Exception _
+         (try (clojure.string/trim (str "DEV-" (:out (sh "git" "describe" "--tags"))))
```
cliffcaseyyet, line 9: i suspect in this case since you are attempting to have a fallback in the case of a non-exact tag that this is where you would want the hash `(sh "git" "rev-parse" "HEAD")`. If we try here at all, that seems more useful than "last tag"

src/main/lrsql/system/openapi.clj
invaliduser: `lrsql/admin/interceptors/openapi.clj`

Makefile
invaliduser: ended up moving away from make

src/main/lrsql/init/git_data.clj
invaliduser: :+1:

src/build/lrsql/build.clj
```diff
@@ -30,13 +30,19 @@
    {:project "deps.edn"
     :aliases [:db-sqlite :db-postgres]}))
 
+(defn write-git-data! []
+  (when-let [version (b/git-process {:git-args "describe --exact-match --tags"})]
+    (b/write-file {:path "resources/lrsql/config/git-details.edn"
```
milt, line 35: I believe this should be `target/classes/lrsql/config/git-details.edn`, and you should fire it _after_ the `b/copy-dir` on line 46. That way we don't write a file that has to be git ignored or the like

milt: note how the copy draws from the source directory and write s to the classes directory. We can skip the extra step by writing directly to the classes directory after that copy

cliffcaseyyet: agree, otherwise this needs a .gitignore right away.

src/main/lrsql/system/webserver.clj
```diff
@@ -101,14 +109,13 @@
...
-               :head-opts
-               {:sec-head-hsts         sec-head-hsts
-                :sec-head-frame        sec-head-frame
-                :sec-head-content-type sec-head-content-type
-                :sec-head-xss          sec-head-xss
-                :sec-head-download     sec-head-download
-                :sec-head-cross-domain sec-head-cross-domain
-                :sec-head-content      sec-head-content}}))
+               :head-opts head-opts}))
+        routes (add-openapi-route
```
cliffcaseyyet, line 113: this is just opinion but it seems weird to me to apply this to the finished routes not just add it in the same clause that adds admin routes.

src/main/lrsql/init/git_data.clj
```diff
@@ -0,0 +1,16 @@
+(ns lrsql.init.git-data
+  (:require [clojure.edn :as edn]
+            [clojure.string]
+            [clojure.java.shell :refer [sh]]))
+
+(defn read-version []
+  (try (edn/read-string (slurp "lrsql/config/git-details.edn"))
```
milt, line 7: This still won't work. You need to wrap that path in `clojure.java.io/resource` to access the classpath.

milt (approved):

src/main/lrsql/system/webserver.clj
cliffcaseyyet: got it

cliffcaseyyet (approved):

PR #403, opened 2024-05-14
AWS Marketplace Configuration

deathtenk:
This adds a separate marketplace cloud formation configuration for LRSQL (with plans to eventually make this the default configuration for LRSQL deploys marketplace or not).


.github/workflows/ami.yml
```diff
@@ -0,0 +1,35 @@
+name: Build AMI
+
+on:
+  push:
+    tags:
+      - 'v*.*.*' # Enforce Semantic Versioning
```
cliffcaseyyet, line 6: have you confirmed what happens if we overwrite a tag? That _is_ something that happens on occasion.

.github/workflows/ami.yml
```diff
@@ -0,0 +1,35 @@
...
+  build_ami:
+    runs-on: ubuntu-latest
+    # These permissions are needed by configure-aws-credentials in order
+    # to interact with GitHub's OIDC Token endpoint.
+    permissions:
+      id-token: write # required to use OIDC authentication
+      contents: read  # required to checkout the code from the repo
+    steps:
+      - name: Install JQ
+        run: sudo apt update && sudo apt install -y jq
```
cliffcaseyyet, line 18: im not seeing where this is utilized

cliffcaseyyet: nevermind, i see it in the sh. Is this usually how we install shell deps?

dev-resources/template/marketplace/0_vpc.yml
```diff
@@ -0,0 +1,224 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
cliffcaseyyet, line 1: should i assume this is identical to old vpc? if so do we need the dupe?

dev-resources/template/marketplace/1_secret.yml
```diff
@@ -0,0 +1,161 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
cliffcaseyyet, line 1: is this whole template just an abstract password generator for use in other templates? It seems like it asks for a path(s), and then CREATES the passwords in SSM? Just kind of skips anyone having to actually go in there i guess...

dev-resources/template/marketplace/ghactions-imagebuilder.yml
```diff
@@ -0,0 +1,39 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
cliffcaseyyet, line 1: so im guessing this one and the next one are more for US not for the actual client deployment?

dev-resources/template/marketplace/3_lrs.yml
```diff
@@ -0,0 +1,937 @@
...
+    Description: Username of seed lrs admin
+  DefaultAdminPass:
+    Description: Initial seed password for lrs admin
+    NoEcho: true
+    Type: String
+  LrsVersion:
+    Description: Version of SQL LRS to pull from Mapping of AMIs
+    Type: String
+    AllowedValues:
+      - v0713
```
cliffcaseyyet, line 58: woof, this is the best they can do with semver? I can see this being an issue down the line...

dev-resources/template/marketplace/3_lrs.yml
```diff
@@ -0,0 +1,937 @@
...
+    Type: Custom::CreateDBUser
+    Condition: IsNotEphemeral
+    Properties:
+      ServiceToken: !GetAtt CreateUserFn.Arn
+      MasterUsername: 
+        Fn::ImportValue: 
+          !Join [":", [!Ref "SecretStackName", "DBMasterUsername"]]
+      MasterPasswordPath:
+        Fn::ImportValue:
+          !Join [":", [!Ref "SecretStackName", "DBMasterUserPasswordPath"]]
```
cliffcaseyyet, line 290: cool

dev-resources/template/marketplace/optional/3_nginx.yml
```diff
@@ -0,0 +1,160 @@
...
+            Statement:
+              - Effect: Allow
+                Action: ["logs:CreateLogGroup", "logs:CreateLogStream", "logs:PutLogEvents"]
+                Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:*"
+              - Effect: Allow
+                Action: ["cloudformation:DescribeStacks"]
+                Resource: "*"
+
+  IngressRuleLambda:
+    Type: AWS::Lambda::Function
```
cliffcaseyyet, line 48: I am unclear on the role of this fn

dev-resources/template/marketplace/0_vpc.yml
deathtenk: it is identical I just duped for consistency, but duping here isn't necessary.

dev-resources/template/marketplace/3_lrs.yml
deathtenk: well the results of the amibuilder are just ami IDs, what im doing here is mapping from semver->region->ami ID. Unfortunately this does mean this template requires manual updates of those mappings for every release.

PR #405, opened 2024-05-20
Empty language maps

kelvinqian00:
- Update the underlying xapi-schema lib version to allow for empty language maps in xAPI Statements
- Despite the above, automatically remove empty maps in xAPI Statements (lang maps, `context`, `result`, and Activity `definition`) before insertion

### DIFF
--- a/src/main/lrsql/util/statement.clj
+++ b/src/main/lrsql/util/statement.clj
@@ -17,14 +17,23 @@
 ;; suggested by the spec:
 ;; https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#requirements-14
 
+(def ^:private dissocable-keys
+  "xAPI Statement properties that can be dissoc'd if the map value is empty.
+   The first three are for Verb, Activity, and Attachment lang maps.
+   The latter three are for top-level Statement or Activity properties.
+   We need to specify in order to allow other maps, e.g. extensions,
+   to remain empty."
+  #{"display" "name" "description" "context" "result" "definition"})
+
 (defn- dissoc-empty-maps
   "Remove all empty maps from `statement`, at all nesting levels."
   [statement]
   (w/postwalk
    (fn [x]
      (if (map-entry? x)
-       (let [v (second x)]
-         (when-not (and (map? v) (empty? v))
+       (let [k (first x)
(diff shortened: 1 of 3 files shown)


cliffcaseyyet (approved):

src/main/lrsql/util/statement.clj
```diff
@@ -9,16 +9,54 @@
...
+
+(defn- dissoc-empty-lang-maps
+  [statement]
+  (cond-> statement
+    ;; Dissoc empty verb display
+    true
+    (update "verb" dissoc-empty-lang-maps*)
+    ;; Dissoc empty object activity name + description
+    (= "Activity" (get-in statement ["object" "objectType"]))
+    (update-in ["object" "definition"]
```
deathtenk, line 34: @kelvinqian00 isn't this too specific? in my dealings with this in the past empty language maps will not pass in if they appear anywhere in a statement. Limiting it to `object.definition` and `attachments` will only allow  cover those two cases, and if the incoming statements have lmaps at other points in the map (for example: `verb.display`), those will not pass through.

@cliffcaseyyet correct me if I'm wrong but we have no way to know if there will or won't be empty lmaps in the statement, and if were too specific we run the risk of making edge cases that fall outside the spec in weird ways.

kelvinqian00: The problem with that is that there are places where empty maps may actually make a semantic difference, specifically in extension values. My first approach was an implementation that generically removed empty maps via `w/postwalk`, caused DATASIM-based tests to fail (since DATASIM likes to put all sorts of values, empty maps included, in extension values), so I changed it down to hyper-specific xAPI property removal.

cliffcaseyyet: Correct me if i'm wrong @kelvinqian00 but isn't that what  `dissoc-empty-lang-maps*` is about? Looks like definition was an edge case where it didn't have a common lmap kw.

kelvinqian00: If you look at the xAPI spec, there are three names for language map properties: `display` (verbs and attachments), `name` (activities), and `description` (activities and attachments). That is what `dissoc-empty-lang-maps*` covers.

PR #406, opened 2024-05-29
Migration from PG VARCHAR to TEXT

cliffcaseyyet:
##Postgres only

Migrate from VARCHAR(255) to unlimited TEXT field to avoid insert errors for ridiculously long IRIs.

Must drop actor_ifi indexes and constraints and rebuild to convert. The rest are trivial

### DIFF
--- a/src/db/postgres/lrsql/postgres/sql/ddl.sql
+++ b/src/db/postgres/lrsql/postgres/sql/ddl.sql
@@ -452,22 +452,17 @@ SELECT 1 FROM information_schema.columns WHERE table_name = 'xapi_statement' AND
 
 -- :name convert-varchars-to-text!
 -- :command :execute
--- :doc Converts all known VARCHAR(255) fields into TEXT fields
+-- :doc Converts all known VARCHAR(255) fields into TEXT fields. Order of execution is critical for ifi constraints
 ALTER TABLE xapi_statement ALTER COLUMN verb_iri TYPE TEXT;
-
 ALTER TABLE statement_to_actor DROP CONSTRAINT actor_fk;
-
 ALTER TABLE actor DROP CONSTRAINT actor_idx;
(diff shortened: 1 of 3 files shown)


src/db/postgres/lrsql/postgres/sql/ddl.sql
```diff
@@ -441,3 +441,41 @@ AND pg_get_constraintdef(oid) LIKE '%ON DELETE CASCADE%'
...
+-- :name convert-varchars-to-text!
+-- :command :execute
+-- :doc Converts all known VARCHAR(255) fields into TEXT fields. Order of execution is critical for ifi constraints
+ALTER TABLE xapi_statement ALTER COLUMN verb_iri TYPE TEXT;
+ALTER TABLE statement_to_actor DROP CONSTRAINT actor_fk;
+ALTER TABLE actor DROP CONSTRAINT actor_idx;
+ALTER TABLE actor ALTER COLUMN actor_ifi TYPE TEXT;
+ALTER TABLE actor ADD CONSTRAINT actor_idx UNIQUE (actor_ifi, actor_type);
+ALTER TABLE statement_to_actor ALTER COLUMN actor_ifi TYPE TEXT;
+ALTER TABLE statement_to_actor ADD CONSTRAINT actor_fk FOREIGN KEY (actor_ifi, actor_type) REFERENCES actor(actor_ifi, actor_type);
```
kelvinqian00, line 462: I would suggest putting a comment before this block explaining why we have to do this ridiculous constraint dropping-and-readding.

milt (approved):

kelvinqian00:
Would it be reasonable to add a test for IFIs, verb IDs, and activity IDs that are over 255 characters long?

kelvinqian00 (approved):

PR #410, opened 2024-06-03
`docker-compose.yml` update

kelvinqian00:
- Change DB's bind mount to a named volume
- Expose DB's `5432` port
- Add a `LRSQL_ALLOW_ALL_ORIGINS` config vars (set it to `false` by default though, to avoid an extra breaking change)


milt (approved):
Woohoo!

PR #419, opened 2024-07-18
Docker remove bench

cliffcaseyyet:
keep container lean.

No need to recut release right now, as its not a priority, just didnt want to forget before the next release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,6 @@ FROM alpine:3.19.1
 ADD target/bundle /lrsql
 ADD .java_modules /lrsql/.java_modules
 
-
 # replace the linux runtime via jlink
 RUN apk update \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -14,6 +14,9 @@ RUN apk update \
         && apk del openjdk11 \
         && rm -rf /var/cache/apk/*
 
+# delete bench utils for leaner container
+RUN rm -rf /lrsql/bench*
```
kelvinqian00, line 18: That simple?

cliffcaseyyet: yeah well there's two things. `bench.jar` and `bench/*.json`. This seems to remove both.

milt (approved):

kelvinqian00 (approved):

PR #421, opened 2024-08-19
[LRS-83] POST with empty list

deathtenk:
This adds a test for POST with empty list

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -45,17 +45,13 @@
   ;; Yet Analytics deps
   
   com.yetanalytics/lrs
-  {#_#_:mvn/version "1.2.19"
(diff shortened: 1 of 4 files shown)


kelvinqian00:
@cliffcaseyyet (Sorry just saw that you responded) Actually I would not close this PR, since I added commits to this branch to bring in the updated schema deps + new tests.

cliffcaseyyet:
> @cliffcaseyyet (Sorry just saw that you responded) Actually I would not close this PR, since I added commits to this branch to bring in the updated schema deps + new tests.

Ah ok, i thought this was the alternative path from what you did.

PR #426, opened 2024-09-24
Tla demo

cliffcaseyyet:


doc/other_demos.md
```diff
@@ -0,0 +1,32 @@
...
+    docker compose up
+
+## TLA Demo
+
+This demo illustrates a configuration similar to the Total Learning Architecture, wherein multiple Noisy LRS instances are feeding a single Transactional LRS. In this demo, three PostgreSQL-backed LRSs will be launched, three LRS Pipe processes will consume their data, and a Transactional LRS will receive the aggregation of that data.
+
+### Run the Docker Stack
+
+    cd dev-resources/tla-demo
+    docker compose up
```
kelvinqian00, line 30: Side note, but I wonder if we should create a separate `demo` folder that's either a subdirectory of `dev-resources` or completely separate from it.

dev-resources/tla-demo/docker-compose.yml
```diff
@@ -0,0 +1,211 @@
...
+  # Noisy LRS 1
+  # - Noisy LRS 1: DB
+  nsy_lrs_1_db:
+    image: postgres:14
+    volumes:
+      - nsy_lrs_1_db_data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: lrsql_user
+      POSTGRES_PASSWORD: lrsql_password
+      POSTGRES_DB: lrsql_db
```
kelvinqian00, line 22: Does this PG DB not have its port exposed? I'm assuming it's because it doesn't need to be access outside the Docker stack.

dev-resources/tla-demo/docker-compose.yml
```diff
@@ -0,0 +1,211 @@
...
+  txn_lrs_db:
+    image: postgres:14
+    volumes:
+      - txn_lrs_db_data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: lrsql_user
+      POSTGRES_PASSWORD: lrsql_password
+      POSTGRES_DB: lrsql_db
+    ports:
+      - "5431:5432"
```
kelvinqian00, line 123: But _here_ the DB port is exposed. Is it just because the previous DB can implicitly be accessed from the outside at port 5432?

dev-resources/tla-demo/docker-compose.yml
```diff
@@ -0,0 +1,211 @@
...
+      LRSQL_DB_NAME: lrsql_db
+      LRSQL_DB_USER: lrsql_user
+      LRSQL_DB_PASSWORD: lrsql_password
+      LRSQL_POOL_INITIALIZATION_FAIL_TIMEOUT: 10000
+      LRSQL_ENABLE_REACTIONS: true
+      LRSQL_ALLOW_ALL_ORIGINS: true
+    restart: always
+
+  # LRSPipe
+  # - LRSPipe: Shared Redis
```
kelvinqian00, line 148: Just to clarify, this is just simple xAPI statement aggregation, with no statement filtering whatsoever, right?

doc/other_demos.md
cliffcaseyyet: yeah, maybe down the line. Right now i've got a client directly looking at it at this address

dev-resources/tla-demo/docker-compose.yml
cliffcaseyyet: right, so this is the transactional and the demo illustrates you could connect an analytics suite to the aggregate, but dont really need to look at the noisy dbs since they're duplicated over.

dev-resources/tla-demo/docker-compose.yml
cliffcaseyyet: yeah, no filters here.

dev-resources/tla-demo/docker-compose.yml
cliffcaseyyet: nope. the noisy LRSs don't really need their DBs examined. Within the docker ecosystem the app can see them just fine, exposing them would just be to the host computer.

dev-resources/tla-demo/docker-compose.yml
kelvinqian00: But then what happens if you make it `"5432:5432"`? Will there be a port clash?

kelvinqian00: Yeah I was confused if there was some mystery port 5432 exposed somewhere else in the docker compose. But that reasoning also makes sense (albeit not obvious at first).

dev-resources/tla-demo/docker-compose.yml
```diff
@@ -0,0 +1,240 @@
+# Example Architecture with Two Noisy LRS' and one Transactional LRS
+
+version: "3.9"
```
milt, line 3: > WARN[0000] /Users/milt/projects/lrsql/dev-resources/tla-demo/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion

milt (approved):
Aside from version warning, works great.

PR #429, opened 2024-10-07
patch commons-io to 2.14.0 to resolve CVE-2024-47554

milt:
patch commons-io to 2.14.0 to resolve CVE-2024-47554

Resolves https://github.com/yetanalytics/lrsql/security/dependabot/17

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -16,6 +16,7 @@
   aero/aero                                {:mvn/version "1.1.6"}
   selmer/selmer                            {:mvn/version "1.12.59"}
   ch.qos.logback/logback-classic           {:mvn/version "1.3.14"}
+  commons-io/commons-io                    {:mvn/version "2.14.0"}
   ;; DB/JDBC deps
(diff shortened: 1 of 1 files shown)


deps.edn
```diff
@@ -16,6 +16,7 @@
   aero/aero                                {:mvn/version "1.1.6"}
   selmer/selmer                            {:mvn/version "1.12.59"}
   ch.qos.logback/logback-classic           {:mvn/version "1.3.14"}
+  commons-io/commons-io                    {:mvn/version "2.14.0"}
```
kelvinqian00, line 19: Is there not an exclusion somewhere?

milt: Don't need it. If you pin a top-level dep, you'll note all the conflicting subdeps in the tree will say `:use-top` and not install

milt: They were much more necessary the way Lein did deps. I think there are still some good cases where we can use them but it just isn't important here.

kelvinqian00 (approved):

PR #430, opened 2024-10-16
Jetty update

kelvinqian00:


kelvinqian00:
We will update to Jetty 12 instead, since Jetty 11 is itself EOL.

PR #433, opened 2024-10-29
[SQL-267] `/admin/verify` endpoint

kelvinqian00:
Add an `/admin/verify` endpoint* that returns a bodyless `204 OK` response when the current JWT is authenticated. This is essentially the `/admin/me` endpoint but without the body. This will be useful for the client to use when polling the server to check if the user is currently authenticated.

Also add `/admin/openapi` and `/admin/status` to the endpoint docs since for some reason they were not included before.

NOTE: name may be subject to change based on suggestions

### DIFF
--- a/src/main/lrsql/admin/interceptors/account.clj
+++ b/src/main/lrsql/admin/interceptors/account.clj
@@ -250,10 +250,10 @@
                :response
                {:status 200 :body result})))}))
 
-(def bodyless
-  "Return a 204 OK response without a body."
+(def no-content
+  "Return a 204 No Content response, without a body."
   (interceptor
-   {:name ::get-bodyless
+   {:name ::get-no-content
     :enter
     (fn get-account [ctx]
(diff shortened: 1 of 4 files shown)


src/main/lrsql/admin/interceptors/account.clj
```diff
@@ -250,6 +250,14 @@
                :response
                {:status 200 :body result})))}))
 
+(def bodyless
```
milt, line 253: While "bodyless" is spooky and Halloween-appropriate (ex. "The Bodyless Horseman"), 204 is called "No Content" and it might be better to source the name from that

kelvinqian00: https://github.com/yetanalytics/lrsql/pull/433/commits/35cfbda14a782d743231ce26618ae3d11e2c8136

milt (approved):

PR #435, opened 2024-11-01
[SQL-268] Admin JWT renewal endpoint

kelvinqian00:
Add a `/admin/account/renew` endpoint and a `LRSQL_JWT_REFRESH_EXP_TIME` config var. The renewal endpoint allows the logged-in user to stay logged in by issuing a new JWT to replace their about-to-be-expired one, while the new config var limits the time when a JWT can be renewed, preventing indefinite logins.


resources/lrsql/config/test/default/webserver.edn
```diff
@@ -4,6 +4,7 @@
  :key-enable-selfie         true
  :jwt-exp-time              3600
  :jwt-exp-leeway            1
+ :jwt-refresh-exp-time      86400
```
cliffcaseyyet, line 7: is that 24h??

cliffcaseyyet (commented):
i think i am not understanding the renewal situation here. Why are we allowing renewal times after expiry times. If it expires you should be logged out and have to log back in.

kelvinqian00:
^ Additional commits to provide necessary env vars to the client

resources/lrsql/config/test/default/webserver.edn
kelvinqian00: Yes, 86400 seconds = 24 hours

cliffcaseyyet (approved):

PR #438, opened 2024-11-18
Activity Definition Upsert

milt:
This PR illustrates a bug with SQL LRS where certain activity definition fields can be "forgotten" when multiple instances of an activity enter the LRS, one without a definition and one with. While the name and description fields appear to be preserved, the `type` field does not.

I've modified `stmt-1` in the lrs tests to have an activity type to show the issue. Note that when `stmt-0` (without a definition) precedes `stmt-1` (with a definition), the resulting activity query is missing the type field. When it is the other way around, `stmt-1` preceding `stmt-0`, the type field shows up.

The issue is resolved by updating the `lrs` lib's `merge-activity` function to allow updates to all definition fields, and to atomically update interaction activity fields.

### DIFF
--- a/src/test/lrsql/lrs_test.clj
+++ b/src/test/lrsql/lrs_test.clj
@@ -84,7 +84,8 @@
                         "zh-CN" "回答了"}}
    "object" {"id"         "http://www.example.com/tincan/activities/multipart"
              "objectType" "Activity"
-             "definition" {"name"        {"en-US" "Multi Part Activity"
+             "definition" {"type"        "http://www.example.com/activity-types/test"
+                           "name"        {"en-US" "Multi Part Activity"
                                           "zh-CN" "多元部分Activity"}
                            "description" {"en-US" "Multi Part Activity Description"
                                           "zh-CN" "多元部分Activity的简述"}}}})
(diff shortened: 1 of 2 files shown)


src/test/lrsql/lrs_test.clj
```diff
@@ -211,231 +212,257 @@
...
+        (lrsp/-store-statements lrs auth-ident [stmt-0] [])
+        ;; Activity was updated between stmt-1 and stmt-0
+        (is (= {:activity
+                {"id"         "http://www.example.com/tincan/activities/multipart"
+                 "objectType" "Activity"
+                 "definition" {"type"        "http://www.example.com/activity-types/test"
+                               "name"        {"en-US" "Multi Part Activity"
+                                              "zh-CN" "多元部分Activity"}
+                               "description" {"en-US" "Multi Part Activity Description"
+                                              "zh-CN" "多元部分Activity的简述"}}}}
```
kelvinqian00, line 460: You should extract the activity here and in the other test case into its own def. Not only does it make the test case more concise, it also emphasizes that the two statement store orders are commutative.

milt: The existing activity query test (and a BUNCH of other tests) just write them out. I personally find it more readable than having to constantly check the top of the NS when working on it. That said, I can just reach into the statement for some of them so I'll do that on the code I touched, but everything else is out of scope.

milt: https://github.com/yetanalytics/lrsql/pull/438/commits/baee73d17490fd98a683f0126ceb7f0a44d2a343

kelvinqian00 (approved):

PR #443, opened 2024-11-27
[SQL-275][SQL-277] Update maven-dependency-submission-action to v4

kelvinqian00:


milt (approved):
Right I forgot, it actually submits deps on any branch (though you can only pull them for master I think), so you can see this working right in the CI list.

PR #462, opened 2025-02-10
[SQL-102] Seed API key flag

kelvinqian00:
Add a `is_seed` column to the `lrs_credential` table and return an optional `seed?` property in the credential body, to denote what is the seed API key pair. It will be used to [notify users if the seed key pair still exists](https://github.com/yetanalytics/lrs-admin-ui/pull/102).


milt (approved):
Seems hygenic

PR #467, opened 2025-03-06
[SQL-283] Authentication for CSV export

kelvinqian00:
Implement authentication for the `GET /admin/csv` endpoint, as HTML `download` attributes cannot have headers, thus making the usual JWT authentication impossible.

We introduce the concept of "one-time JWTs" that have "one-time IDs." These one-time JWTs are stored in the `jwt_blocklist` SQL table (added in v0.8.0), where if the `one_time_id` column is not null then they are ignored during the query for blocked JWTs. Thus, these JWTs can easily be blocked/revoked by updating `one_time_id` to be null, which happens once they are validated. These one-time JWTs can be added as a `token` parameter for any GET endpoint, this case the CSV download endpoint.

### DIFF
--- a/src/db/postgres/lrsql/postgres/record.clj
+++ b/src/db/postgres/lrsql/postgres/record.clj
@@ -80,7 +80,8 @@
       (add-statement-to-actor-cascading-delete! tx))
     (when (some? (query-varchar-exists tx))
       (convert-varchars-to-text! tx))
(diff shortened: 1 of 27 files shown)


cliffcaseyyet (approved):
This seems like it should work well. I was questioning the table reuse/repurpose but i think its been pretty well handled.

PR #514, opened 2026-06-08
Makefile: Reference admin index.html explicitly

lmkz:
After cloning repo and performing "make ephemeral", web server returns an error as admin UI SPA files are not downloaded. /public/admin folder is initially present but empty. Changed Makefile to instead check for presence of index.html.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ LRS_ADMIN_UI_LOCATION ?= https://github.com/yetanalytics/lrs-admin-ui/releases/d
 LRS_ADMIN_ZIPFILE ?= lrs-admin-ui-${LRS_ADMIN_UI_VERSION}.zip
 
 # Get the admin UI SPA release from GitHub
-resources/public/admin:
+resources/public/admin/index.html:
 	curl -L ${LRS_ADMIN_UI_LOCATION} -o ${LRS_ADMIN_ZIPFILE}
 	mkdir -p resources/public/admin
 	unzip ${LRS_ADMIN_ZIPFILE} -d resources/public/admin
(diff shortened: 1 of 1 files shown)


cliffcaseyyet:
Hi @lmkz ,

We appreciate the contribution. I cannot seem to reproduce the issue you referenced. Can you provide reproduction steps? I did: 

- clean clone of main 
- ran nothing but `make ephemeral`
- it properly unzipped the admin UI.

Your fix seems harmless but I am just trying to understand the condition in which this would occur. Is it possible the `public/admin` files were deleted by something outside a make target?

Thanks,
Cliff

lmkz:
Hi @cliffcaseyyet that's basically what I did, but the admin folder content didn't exist. When I get some time I'll try again from scratch with a fresh clone and report back here. Thanks.

lmkz:
Hello @cliffcaseyyet my apologies, I just tried that with a fresh clone and it all worked perfectly. Strange as I did exactly the same steps as previously. Perhaps last time I was missing some dependency which meant the admin folder got created but not its content on the first run (Makefile lines 10-14)? That seems most likely in hindsight, and make clean would've probably solved that.

cliffcaseyyet:
@lmkz Ok, well as I said the PR is pretty harmless from what I can tell so I'll get some other opinions on this end and maybe we'll merge it anyway on the off chance someone deletes the files but *not* the folders. I mainly just wanted to check there wasn't some condition we weren't thinking of. The one case I can think of is:

- you have a machine that doesn't have `unzip`
- You run it and it fails
- then you go get unzip via package manager
- then rerun and the folder has been created so it skips that part

Thanks,
Cliff

lmkz:
@cliffcaseyyet I would say that is highly likely what happened. Thanks!

cliffcaseyyet (approved):

PR #522, opened 2026-08-11
bulk doc atomic delete with serializable txns

milt:
Compare to the observed keys approach here: https://github.com/yetanalytics/lrsql/pull/523


milt:
We're going to go with the observed-keys approach in #523

